PR #6, opened 2019-07-18
Fix python buildpack on OSX Mojave

nictuku:
Background here:

https://forum.yourbase.io/t/modulenotfounderror-no-module-named-pyexpat/12

### DIFF
--- a/buildpacks/python.go
+++ b/buildpacks/python.go
@@ -139,8 +139,8 @@ func (bt PythonBuildTool) InstallPlatformDependencies() error {
 				cmd.Run()
 			}
 		}
-		// Thanks to pablopunk, here:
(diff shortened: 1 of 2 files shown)


buildpacks/python.go
```diff
@@ -139,6 +139,12 @@ func (bt PythonBuildTool) InstallPlatformDependencies() error {
 				cmd.Run()
 			}
 		}
+		// Thanks to henriquebastos, here:
+		// https://github.com/pyenv/pyenv/issues/1066#issuecomment-504710495
+		if strings.HasPrefix(gi.Core, "18.6") {
```
johnewart: I feel like maybe these are things lots of build packs will need... 🤔 but it's fine here for now

PR #7, opened 2019-07-18
Support installing homebrew packages as dependencies

johnewart:
Also provide a checksum for the build manifest's build dependencies and a -deps-only flag for build

### DIFF
--- a/buildpacks/homebrew.go
+++ b/buildpacks/homebrew.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/matishsiao/goInfo"
 	. "github.com/yourbase/yb/plumbing"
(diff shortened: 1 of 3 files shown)


buildpacks/homebrew.go
```diff
@@ -44,15 +61,51 @@ func (bt HomebrewBuildTool) InstallDir() string {
...
 	case "darwin":
-		return bt.InstallDarwin()
+		err = bt.InstallDarwin()
 	case "linux":
-		return bt.InstallLinux()
+		err = bt.InstallLinux()
 	default:
-		fmt.Printf("Unsupported platform: %s\n", gi.GoOS)
-		return nil
+		err = fmt.Errorf("Unsupported platform: %s\n", gi.GoOS)
```
nictuku: no need for newline in Errorf, same below

nictuku: ..and better not have it

types/build_manifest.go
```diff
@@ -1,9 +1,23 @@
 package types
 
 import (
+	"bytes"
+	"crypto/sha256"
 	"fmt"
 )
 
+const PKG_CHECKSUM_LENGTH = 12
```
nictuku: PkgChecksumLength, constants are usually CamelCase.

buildpacks/homebrew.go
johnewart: Fixed -- conversion from Printf artifact

types/build_manifest.go
johnewart: Done

PR #11, opened 2019-07-23
Replace / with _ in the container name when creating containers

johnewart:


johnewart:
Wouldn't hurt -- there is a list of characters that are prohibited but I believe they match the list for images (other than the /) so it should be mostly safe

PR #12, opened 2019-07-23
anaconda buildpack for Python3

beholders-eye:
My first buildpack, I tested it with https://github.com/project-kotinos/lneuhaus___pyrpl/, and it worked.

### DIFF
--- a/buildpacks/anaconda.go
+++ b/buildpacks/anaconda.go
@@ -1,0 +1,127 @@
+package buildpacks
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	. "github.com/yourbase/yb/plumbing"
+	. "github.com/yourbase/yb/types"
+)
+
+type AnacondaBuildTool struct {
+	BuildTool
+	version string
+	spec    BuildToolSpec
+}
+
+// Fixed for Python3, later we'll create some support for Python2
+var ANACONDA_DIST_MIRROR = "https://repo.continuum.io/miniconda/Miniconda3-{{.Version}}-Linux-x86_64.sh"
+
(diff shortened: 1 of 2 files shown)


johnewart:
Hey this will work for one Adam was doing — I was just going to write one today so that’s awesome! 🤘🏻

buildpacks/anaconda.go
```diff
@@ -0,0 +1,127 @@
...
+)
+
+type AnacondaBuildTool struct {
+	BuildTool
+	version string
+	spec    BuildToolSpec
+}
+
+// Fixed for Python3, later we'll create some support for Python2
+var ANACONDA_DIST_MIRROR = "https://repo.continuum.io/miniconda/Miniconda3-{{.Version}}-Linux-x86_64.sh"
```
johnewart: Does this work for other systems or just Linux? If others can we make the URL template take those properties?

buildpacks/anaconda.go
```diff
@@ -0,0 +1,127 @@
...
+	. "github.com/yourbase/yb/plumbing"
+	. "github.com/yourbase/yb/types"
+)
+
+type AnacondaBuildTool struct {
+	BuildTool
+	version string
+	spec    BuildToolSpec
+}
+
```
johnewart: We should probably name this anaconda3 if the version doesn’t follow Python

buildpacks/anaconda.go
beholders-eye: I just forgot to move the variable up, I had it right. Fixing it...

johnewart:
👍🏻

PR #14, opened 2019-07-24
YML root --> workdir

beholders-eye:
Using "root" to specify a different working directory is confusing, so proposing, as @nictuku mentioned, this fix. It would break old `.yourbase.yml` files.

### DIFF
--- a/cli/build.go
+++ b/cli/build.go
@@ -415,9 +415,9 @@ func RunCommands(config BuildConfiguration) ([]CommandTimer, error) {
 			//os.Chdir(dir)
 			targetDir = dir
 		} else {
-			if target.Root != "" {
-				fmt.Printf("Build root is %s\n", target.Root)
-				targetDir = filepath.Join(targetDir, target.Root)
+			if target.WorkDir != "" {
+				fmt.Printf("Build root is %s\n", target.WorkDir)
+				targetDir = filepath.Join(targetDir, target.WorkDir)
 			}
 
 			if sandboxed {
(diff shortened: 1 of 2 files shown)


cli/build.go
```diff
@@ -415,9 +415,9 @@ func RunCommands(config BuildConfiguration) ([]CommandTimer, error) {
 			//os.Chdir(dir)
 			targetDir = dir
 		} else {
-			if target.Root != "" {
-				fmt.Printf("Build root is %s\n", target.Root)
-				targetDir = filepath.Join(targetDir, target.Root)
+			if target.WorkDir != "" {
+				fmt.Printf("Build root is %s\n", target.WorkDir)
```
nictuku: s/Build root is/Workdir is/

beholders-eye:
Rebased it and fix the status message, as Yves proposed. I don't think we should support both configs, it could get more confusing, IMO.

nictuku:
In general, if we have important users of a certain syntax, being backwards compatible as a transition would be the right thing to do. We would show a warning saying the old syntax is going away.

In this case, it's possible that nobody is using root so it doesn't matter. But we didn't check, did we?

beholders-eye:
No we didn't.

johnewart:
Speaking of this we should devise a way to track what people are using in their configs at some point (maybe just keep heuristics on each part of the config is present, how many, etc.) like the equivalent of browser heat maps to see where people are clicking

beholders-eye:
Probably a whole new component, don't you think? We could process all YML created on Dispatcher or the Build Agent itself, and send some kind of aggregation to this new component/server that would show a syntax heat map. Did I get it right?

johnewart:
I was just thinking earlier that we should abstract a number of the struct details as functions (possibly as an interface) so we can do things like support both for a period of time if we transition. Right now it's too tightly coupled to the details of the config struct and we will eventually want to version the config files

beholders-eye:
All that being said I'll close this PR, and we'll plan a better way to fix this.

PR #16, opened 2019-07-25
Configuration tweaks

johnewart:
Support a YOURBASE_PROFILE environment variable, consolidate config into a `config` package, clean up a few things related to config.

### DIFF
--- a/cli/login.go
+++ b/cli/login.go
@@ -12,8 +12,9 @@ import (
 	"os"
 	"strings"
 
-	. "github.com/yourbase/yb/plumbing"
 	. "github.com/yourbase/yb/types"
+
+	ybconfig "github.com/yourbase/yb/config"
 )
 
 type LoginCmd struct {
@@ -30,7 +31,7 @@ func (p *LoginCmd) SetFlags(f *flag.FlagSet) {
 
 func (p *LoginCmd) Execute(_ context.Context, f *flag.FlagSet, _ ...interface{}) subcommands.ExitStatus {
 	reader := bufio.NewReader(os.Stdin)
-	fmt.Println("Open up https://app.yourbase.io/user/apitoken and then paste the token here.")
+	fmt.Println(fmt.Sprintf("Open up %s and then paste the token here.", ybconfig.ManagementUrl("user/apitoken")))
 	fmt.Println()
 
 	fmt.Print("Token: ")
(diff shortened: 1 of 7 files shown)


cli/login.go
```diff
@@ -30,7 +31,7 @@ func (p *LoginCmd) SetFlags(f *flag.FlagSet) {
 
 func (p *LoginCmd) Execute(_ context.Context, f *flag.FlagSet, _ ...interface{}) subcommands.ExitStatus {
 	reader := bufio.NewReader(os.Stdin)
-	fmt.Println("Open up https://app.yourbase.io/user/apitoken and then paste the token here.")
+	fmt.Println(fmt.Sprintf("Open up %s and then paste the token here.", ybconfig.ManagementUrl("user/apitoken")))
```
nictuku: can you do a single `fmt.Printf`? This is kinda hard to read.

config/settings.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	profile := YourBaseProfile()
+
+	switch profile {
+	case "staging":
+		return "https://api.staging.yourbase.io"
+	case "development":
+		return "http://localhost:5000"
+	default:
```
nictuku: Maybe it's more helpful / user friendly to be more strict. I would do:

case "production":
case " ":
  return prod-stuff, nil

default:
   return ''
", error

That sadly means changing all call paths of apiBaseURL to take an error value too but it could save some frustration for someone trying to understand why the "dev" profile (or another invalid value) doesn't work.

config/settings.go
```diff
@@ -0,0 +1,114 @@
...
+
+	return managementURL
+}
+
+func UserToken() (string, error) {
+	token, exists := os.LookupEnv("YB_USER_TOKEN")
+	if !exists {
+		if token, err := GetConfigValue("user", "api_key"); err != nil {
+			return "", fmt.Errorf("Unable to find YB token in config file or environment.\nUse yb login to fetch one, if you already logged in to https://app.yourbase.io")
+		} else {
```
nictuku: you can remove the `else` here and below because there are `returns` already. It makes the code more linear :)

johnewart:
As you wish :smile:

cli/patch.go
```diff
@@ -221,10 +217,6 @@ func commitSet(r *git.Repository) map[string]bool {
 		return nil
 	})
 
-	if err != nil {
-		fmt.Printf("Error printing: %v\n", err)
```
beholders-eye: Cool, I was wondering if this should go or not.

cli/remote_build.go
```diff
@@ -451,45 +393,6 @@ func fetchProject(urls []string) (*Project, error) {
 	return &project, nil
 }
 
-func fetchUserEmail() (string, error) {
```
beholders-eye: Another thing I was wondering. I didn't find where it was being used. Do we really won't use such information even in the near future?

PR #17, opened 2019-07-31
Use Anaconda for Python build pack

johnewart:
More reliable and faster than building from source -- thoughts?

### DIFF
--- a/buildpacks/python.go
+++ b/buildpacks/python.go
@@ -3,12 +3,7 @@ package buildpacks
 import (
 	"fmt"
 	"os"
-	"os/exec"
 	"path/filepath"
-	"strings"
-
-	"github.com/matishsiao/goInfo"
-	"gopkg.in/src-d/go-git.v4"
 
 	. "github.com/yourbase/yb/plumbing"
 	. "github.com/yourbase/yb/types"
@@ -19,6 +14,10 @@ type PythonBuildTool struct {
 	version string
 	spec    BuildToolSpec
 }
+
+var ANACONDA_URL_TEMPLATE = "https://repo.continuum.io/miniconda/Miniconda{{.PyNum}}-{{.Version}}-{{.OS}}-{{.Arch}}.{{.Extension}}"
+
+const AnacondaToolVersion = "4.7.10"
 
 func NewPythonBuildTool(toolSpec BuildToolSpec) PythonBuildTool {
 	tool := PythonBuildTool{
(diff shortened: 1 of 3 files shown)


johnewart:
Also a minor bug fix for `exec` so that if the dependency setup fails then `exec` fails

buildpacks/python.go
```diff
@@ -33,119 +32,120 @@ func (bt PythonBuildTool) Version() string {
...
+			return err
 		}
-	}
 
-	if _, err := os.Stat(pythonVersionDir); err == nil {
-		fmt.Printf("Python %s installed in %s\n", bt.Version(), pythonVersionDir)
-	} else {
-		os.Setenv("PYENV_ROOT", pyenvDir)
-		PrependToPath(filepath.Join(pyenvDir, "bin"))
+		// TODO: Windows
```
beholders-eye: I think we could trigger `Linux for Windows Subsystem` to test this, we can open another issue for that.

buildpacks/python.go
```diff
@@ -33,119 +32,120 @@ func (bt PythonBuildTool) Version() string {
...
+	if _, err := os.Stat(envDir); err == nil {
+		fmt.Printf("environment installed in %s\n", envDir)
+	} else {
+		currentPath := os.Getenv("PATH")
+		newPath := fmt.Sprintf("PATH=%s:%s", filepath.Join(condaDir, "bin"), currentPath)
+		setupDir := bt.spec.PackageDir
+		condaBin := filepath.Join(condaDir, "bin", "conda")
+
+		for _, cmd := range []string{
+			fmt.Sprintf("%s config --set always_yes yes --set changeps1 no", condaBin),
```
beholders-eye: I'm not sure if this commands should be used every time to setup Anaconda/Miniconda

johnewart: It can't hurt; it does ensure that the config is properly set

johnewart: Miniconda does provide a python binary, we then install the requested python in an environment and use that.

buildpacks/python.go
johnewart: Not everyone uses WSL, we should make sure we have an install / setup that works on pure Windows :)

nictuku:
> If miniconda does provide a `python` executable, it would be a good substitute for `pyenv`. But I think it would be good to test more then one Python version, like 3.7.4 or 3.6.9.

Meta-point: Was this comment meant to block the PR? It doesn't seem actionable. When reviewing PRs, let's try to unblock the author ASAP. 👍

PR #18, opened 2019-07-31
remotebuild: respect the existing remote in the local fetch

nictuku:
Pair-programmed with @beholders-eye 

fixes microclusters/issues#212 

But there's a big TODO that should be dealt with.  But..

image

We need to find which local remote was picked.

We should fix that before merging :-)

### DIFF
--- a/cli/remote_build.go
+++ b/cli/remote_build.go
@@ -170,7 +170,7 @@ func (p *RemoteCmd) Execute(_ context.Context, f *flag.FlagSet, _ ...interface{}
 
 	cloneDir, err := ioutil.TempDir("", "yb-clone-")
 	if err != nil {
-		fmt.Printf("Unable to create clone dir to fetch and checkout remote repository '%v': %v\n", project.Repository, err)
+		fmt.Printf("Unable to create clone dir to fetch and checkout remote repository '%v': %v\n", project.LocalRepoRemote, err)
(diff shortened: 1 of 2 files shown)


cli/remote_build.go
```diff
@@ -389,6 +389,10 @@ func fetchProject(urls []string) (*Project, error) {
 	if err != nil {
 		return nil, err
 	}
+	// TODO: Use the correct URL rather than the first one.
```
beholders-eye: Have the API return the original remote(s) URL(s) sent at the beginning, for yb to clone it locally as the user already does.

types/types.go
```diff
@@ -79,11 +79,12 @@ type BuildTarget struct {
...
-	Label       string `json:"label"`
-	Description string `json:"description"`
-	Repository  string `json:"repository"`
-	OrgSlug     string `json:"org_slug"`
+	Id              int    `json:"id"`
+	Label           string `json:"label"`
+	Description     string `json:"description"`
+	Repository      string `json:"repository"`
+	OrgSlug         string `json:"org_slug"`
+	LocalRepoRemote string `json:"local_repo_remote"`
```
beholders-eye: The API should return this `local_repo_remote` field/property

beholders-eye:
#25 has this fix too, thanks @nictuku

PR #19, opened 2019-07-31
Keep alive between yb and dispatcher

nictuku:
For microclusters/issues#205

It doesn't solve it. Either it's missing the dispatcher `pong` or this is not the real cause of the problem.


nictuku:
The pong handler was removed here:

https://github.com/microclusters/dispatcher/commit/3353f7d29a7314b5f8ed8b88de19355c5080050c

I'll check with John to understand if we can put it back.

nictuku:
I'm going to close this in favor of just using the branch https://github.com/yourbase/yb/tree/keep-alive. (It's adding noise to Toast)

PR #22, opened 2019-08-08
Add support for quotes in commands and some exec helpers

johnewart:
Commands with quotes are supported; also adding some exec helpers to be quiet and show progress as well as log; we might consider consolidating these so they're consistent in behavior

### DIFF
--- a/plumbing/util.go
+++ b/plumbing/util.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


johnewart:
Addresses #207

plumbing/util.go
```diff
@@ -48,6 +55,90 @@ func ExecToStdout(cmdString string, targetDir string) error {
...
+			}
+		}
+	}()
+
+	go func() {
+		for {
+			select {
+			default:
+			case <-dotchan:
+				fmt.Printf(".")
```
beholders-eye: It may look better if we only show 5 dots at max, after 5 we should delete all dots and start printing them again, inplace.

johnewart: I thought about that but, in particular with remote logs being sent back we don't have control over the terminal directly and so it makes the streamed logs look funny. We could increase the time between dots maybe; thoughts?

beholders-eye: Yeah, maybe 10 seconds

PR #24, opened 2019-08-14
Support downloading pre-build Ruby binaries we host

johnewart:


nictuku:
@johnewart  is this ready for review?

PR #25, opened 2019-08-14
Remote Build improvements

beholders-eye:
Now you can invoke `yb remotebuild -committed` if you just want to send local committed changes, the default behavior is to also generate a patch including untracked (but git ignored) files and staged files.

`yb remotebuild -patch-path <patch.diff>` saves the generated patch, so `yb patch` was removed.

Also introduced `yb -dry-run` which disables submitting the build, it just do everything related to the patch generation. (So flag `-patch-path` should be used together with this one).

Brought @nictuku PR (#18 ) to this one, which would work best if microclusters/yourbase_api#20 is merged too.


beholders-eye:
@johnewart PTAL, I've also fixed how we handle git remotes, created a parser to check and save them

types/gitremote.go
```diff
@@ -0,0 +1,156 @@
...
+		case 3:
+			str = fmt.Sprintf(template, auth, r.Domain, r.Path)
+		case 4:
+			str = fmt.Sprintf(template, r.Protocol, auth, r.Domain, r.Path)
+		}
+	}
+	return
+}
+
+func parseRemote(s string) (token, user, password, domain, path, protocol string, t RemoteType) {
```
nictuku: Hmm I really think we should add a unit test for this, using this style:

https://github.com/golang/go/wiki/TableDrivenTests

cli/remote_build.go
```diff
@@ -156,112 +165,222 @@ func (p *RemoteCmd) Execute(_ context.Context, f *flag.FlagSet, _ ...interface{}
...
 	return subcommands.ExitSuccess
 }
 
+func commitTempChanges(w *git.Worktree) (latest plumbing.Hash, err error) {
+	latest, err = w.Commit(
+		"YourBase remote build",
+		&git.CommitOptions{
+			Author: &object.Signature{
+				Name:  "YourBase",
+				Email: "robot@yourbase.io",
```
johnewart: We should make a TODO to get this from their config or from the API so that people know who's building

PR #26, opened 2019-08-15
validate YML configurations

beholders-eye:
It tries two things, just as `yb build` does:

1. Check for the existence and correctness of `.yourbase.yml`
2. If it doesn't exist, look for the Workspace YML on parent directories


cli/checkconfig.go
```diff
@@ -0,0 +1,62 @@
...
+			fmt.Printf("Error loading package '%s': %v\n", pkgName, err)
+			return subcommands.ExitFailure
+		}
+		targetPackage = pkg
+	} else {
+
+		workspace, err := LoadWorkspace()
+
+		if err != nil {
+			fmt.Printf("No package here, and no workspace, nothing to check!")
```
nictuku: Do you think a completely new user of yourbase is going to understand this message? Is there a way we can be more helpful and clear?

Even I don't know what workspace and project mean in this context. How about:

"Could not find .yourbase.yml file. Try running yb in the repository root or writing the YAML config file if it missing. See https://docs.yourbase.io"

cli/checkconfig.go
```diff
@@ -0,0 +1,62 @@
...
+			return subcommands.ExitFailure
+		}
+
+		pkg, err := workspace.TargetPackage()
+		if err != nil {
+			fmt.Printf("Can't load workspace's target package: %v\n", err)
```
nictuku: fmt.Printf("Can't load workspace's target package: %v. Please check that it is defined in .yourbase.yml\n", err)

cli/checkconfig.go
beholders-eye: Good call

beholders-eye: Actually, this entire else block os for loading the Workspace, not directly the .yourbase.yml file. But I get your point, I'll improve the user messages.

cli/checkconfig.go
beholders-eye: Actually this depends on loading the `config.yml` from the upper Workspace directory

beholders-eye:
@nictuku PTAL

beholders-eye:
@nictuku PTAL

PR #27, opened 2019-08-15
README: how to use development channel

nictuku:


README.md
```diff
@@ -45,9 +45,17 @@ The primary features of the YB tooling are:
 
 # Documentation 
 
-We are working on publishing documentation but for now you can look at the
-`.yourbase.yml` file included in this repository or look in the Wiki for some
-examples of how to get started using the tool.
+See http://docs.yourbase.io.
```
beholders-eye: We could add: 

To build this package (repository) based on instructions defined inside `.yourbase.yml`

`yb build`

You can insert an prefix to all YML defined commands by using `-exec-prefix`:

`yb build -exec-prefix "time "`

You can force bypass running inside a container, even if `.yourbase.yml` defines one, by using `-no-container`:

`yb build -no-container`

You can only install the dependencies without actually building by using `-deps-only`:

`yb build -deps-only`

To execute the package, starting it's listening services, workers or other processes, also described inside `.yourbase.yml`:

`yb exec`

You can use another set of environment variables by using `-e`:

`yb exec -e no-tests`

To get local changes and send to the YourBase infrastructure for building this package remotely, that needs you to login first using:

`yb login`

then:

`yb remotebuild`

You can save the generated patch by using `-patch-path` and prevent from really submitting it by using `-dry-run`:

`yb remotebuild -patch-path locally-yb.diff -dry-run`

You can define an ancestor common base commit that should exist in the remote repo, by using `-base-commit`:

`yb remotebuild -base-commit ef89da1`

You can define a branch, that should exist in the remote repo, to compare to by using `-branch`:

`yb remotebuild -branch dev`

You can switch to another target defined inside `.yourbase.yml` by using `-target`:

`yb remotebuild -target special-very-looong-running-tests`

To show the installed CLI tool's version

`yb version`

To generate an archive for the package:

`yb package`

You can create the package for another target by using `-target`:

`yb package -target big-one`

To syntax check `.yourbase.yml`:

`yb checkconfig`

nictuku: I think you should do this in a separate PR, not block this one, right?

I think having a more comprehensive README would be helpful. Can we add your suggestions to the official help of each command, in the Go code? Then we can show the full `yb help` output here?

But, again, I think that should be a separate PR.

nictuku:
@beholders-eye thanks man but can you make a PR with your proposed (good) change? That is somewhat unrelated to this PR and it's blocking me from merging the change in the first place.

nictuku:
I'm sorry but I'm going to forcebly resolve your comments.  Like I said, I need to get this PR merged and your comment is about introducing a new change.

README.md
beholders-eye: Nice, I'll do it.

PR #43, opened 2019-09-18
Quicker remotebuild

beholders-eye:
Removed the step where we cloned the remote in memory and compared it to the local work tree.

I'll go back to the unified patch apply utility in the future, it should be a good replacement for `patch -i` we use now in the build agent. I'll left that code there as reference.

I inserted timings and a progress spinner to the remote build. It measures how much time it takes to generate the patch and to submit the build to the API+CI+BuildAgent. Right before starting the build itself, we print those timing info.


cli/workspace.go
```diff
@@ -77,8 +77,6 @@ func (w *workspaceLocationCmd) Execute(_ context.Context, f *flag.FlagSet, _ ...
 		fmt.Println(ws.Root()) // No logging used, because this can be used by scripts
 		return subcommands.ExitSuccess
 	}
-
```
beholders-eye: Unreachable code.

plumbing/patch_handle.go
```diff
@@ -0,0 +1,154 @@
+package plumbing
```
beholders-eye: I'll finish it in the near future, for use inside the build agent.

plumbing/progress.go
```diff
@@ -0,0 +1,247 @@
+// This came from https://raw.githubusercontent.com/BTBurke/clt/7f5151b6bef1b0da4b03f965b47cbe2de87a3ac0/progress.go
```
beholders-eye: Maybe we could consider using BTBurke/clt for everything inside our CLI...

types/worktreesave.go
```diff
@@ -0,0 +1,66 @@
+package types
```
beholders-eye: I created this just for a foreseeable scenario where `git reset` from go-git doesn't really work and we may destroy our customer's work when trying to remote build.

cli/build.go
```diff
@@ -117,7 +119,7 @@ func (b *BuildCmd) Execute(_ context.Context, f *flag.FlagSet, _ ...interface{})
 		// Setup dependencies
 		containers := target.Dependencies.Containers
 
-		if len(containers) > 0 {
+		if len(containers) > 0 && !b.NoSideContainer {
```
beholders-eye: If we want to ignore side containers completly... That was what gave that Kotinos bug today.

cli/remote_build.go
```diff
@@ -348,13 +342,17 @@ func (p *RemoteCmd) Execute(_ context.Context, f *flag.FlagSet, _ ...interface{}
...
+		err = fmt.Errorf("Needs a worktree and a commit object")
+		return
+	}
 	latest, err = w.Commit(
 		"YourBase remote build",
 		&git.CommitOptions{
 			Author: &object.Signature{
-				Name:  "YourBase",
-				Email: "robot@yourbase.io",
+				Name:  c.Author.Name,
```
beholders-eye: You asked me that a loooooong time ago :laughing:

beholders-eye:
Added the commits from new-release-script branch, I'll release a new preview version to use in the Preview env

beholders-eye:
@johnewart, I have this change sitting locally, because it became a great number of checks :smile: 
```
diff --git a/.yourbase.yml b/.yourbase.yml
index 76f5b93..331eb59 100644
--- a/.yourbase.yml
+++ b/.yourbase.yml
@@ -57,7 +57,7 @@ ci:
 
     - name: preview_release
       build_target: preview_release
-      when: branch IS NOT 'master'
+      when: branch IS NOT 'master' AND tagged IS false
 
     - name: release
       build_target: release
```

cli/remote_build.go
johnewart: :+1:

cli/best_ancestor.go
```diff
@@ -0,0 +1,105 @@
...
+	*/
+	c = -1
+
+	ref, err := r.Head()
+	if err != nil {
+		log.Errorf("%v", err)
+		return
+	}
+
+	remoteBranch := findRemoteBranch(ref, r)
```
johnewart: We talked about this outside of this diff -- there's a minor issue with multiple remotes so we should only use github ones for the moment. Long term we should probably negotiate a remote that's shared between the user and the CI system (i.e what the github hack will do but generalized)

beholders-eye:
@johnewart PTAL

beholders-eye:
Included a log entry showing complete git remote information.

beholders-eye:
> @johnewart, I have this change sitting locally, because it became a great number of checks smile
> 
> ```
> diff --git a/.yourbase.yml b/.yourbase.yml
> index 76f5b93..331eb59 100644
> --- a/.yourbase.yml
> +++ b/.yourbase.yml
> @@ -57,7 +57,7 @@ ci:
>  
>      - name: preview_release
>        build_target: preview_release
> -      when: branch IS NOT 'master'
> +      when: branch IS NOT 'master' AND tagged IS false
>  
>      - name: release
>        build_target: release
> ```

I committed and pushed this

johnewart:
I think we're good here; merging this is blocked on a minor change to the buildagent so we don't save patched images, which I'll get to today, and then we can merge this :+1:

beholders-eye:
John, I started doing that in the agent already

beholders-eye:
PTAL @johnewart

johnewart:
I'm going to squash and merge this in the super-rare chance we need to revert it since testing seems to work well and then we can make changes going forward :+1:

PR #54, opened 2019-10-25
Download from correct channel

johnewart:
* Auto-determine the channel to download in the containers based on the channel baked into the CLI
* Perform a HEAD request to see if the local cached download matches the remote file size (fixes corrupt cached files)
* Auto-detect development mode (i.e upload the local binary being executed into the container) without needing to set an environment variable

### DIFF
--- a/cli/build.go
+++ b/cli/build.go
@@ -13,6 +13,7 @@ import (
 	"os/signal"
 	"os/user"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"time"
 
@@ -30,6 +31,8 @@ import (
 const TIME_FORMAT = "15:04:05 MST"
 
 type BuildCmd struct {
+	Channel          string
+	Version          string
 	ExecPrefix       string
 	NoContainer      bool
 	NoSideContainer  bool
@@ -184,7 +187,7 @@ func (b *BuildCmd) Execute(_ context.Context, f *flag.FlagSet, _ ...interface{})
 		}
 
 		// Do our build in a container - don't do the build locally
-		buildErr := BuildInsideContainer(target, containerOpts, buildData, b.ExecPrefix, b.ReuseContainers)
+		buildErr := b.BuildInsideContainer(target, containerOpts, buildData, b.ExecPrefix, b.ReuseContainers)
 		if buildErr != nil {
 			log.Errorf("Unable to build %s inside container: %v", target.Name, buildErr)
 			return subcommands.ExitFailure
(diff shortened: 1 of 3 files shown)


cli/build.go
```diff
@@ -704,17 +697,14 @@ func (b BuildData) EnvironmentVariables() []string {
...
+	// Stable version stable URL
+	downloadUrl := "https://bin.equinox.io/c/3fw6ZrH8UFD/yb-stable-linux-amd64.tgz"
 	currentPath, _ := filepath.Abs(".")
 	tmpPath := filepath.Join(currentPath, ".yb-tmp")
-	MkdirAsNeeded(tmpPath)
 	ybPath := filepath.Join(tmpPath, "yb")
-	if _, err := os.Stat(ybPath); err == nil {
-		return ybPath, nil
-	}
+	MkdirAsNeeded(tmpPath)
```
herbertmd: Did you mean to change this to ybPath, since you moved the call to after setting that variable?

cli/build.go
```diff
@@ -431,12 +438,6 @@ func BuildInsideContainer(target BuildTarget, containerOpts BuildContainerOpts,
 		} else {
 			localYbPath = p
 		}
-		// After downloading YB and uploading it to the container, we should delete it locally
```
beholders-eye: Without removing `.yb-tmp/` we'll have to add this to every `.gitignore` that uses the CLI... Why did you remove this?

cli/build.go
beholders-eye: Actually tmpPath == path to the temporary directory that will keep the yb binary; ybPath == tmpPath + "/yb", right?

cli/build.go
```diff
@@ -726,8 +716,3 @@ func DownloadYB() (string, error) {
 
 	return ybPath, nil
 }
-
-func DeleteLocalYB() error {
```
beholders-eye: I don't agree with this. Getting yb from the correct channel and auto-detecting devmode don't invalidate the need to cleanup this temporary directory. Maybe we could use a real temporary directory by using `os.TempDir()` (https://godoc.org/os#TempDir), as most modern O.S. clean them up at boot or shutdown.

johnewart: Seems reasonable -- I changed it to use our cache dir, also checked to see if the binary checksum matches the known good one before bothering to try re-downloading

cli/build.go
johnewart: Changed, see below / commit comment

cli/build.go
johnewart: Ridai is correct -- tmpPath is where we un-archive the binary to, so we need to make it before we explode the archive

PR #51, opened 2019-10-11
Remote build public projects (#80)

beholders-eye:
https://app.clubhouse.io/yourbaseio/story/80/yb-remotebuild-should-work-even-if-project-not-there-yet

Updates the API response expectations; initially supports building public repos (if GH would work)

PR #55, opened 2019-10-25
add url list when project fetch fails

beholders-eye:
https://app.clubhouse.io/yourbaseio/story/827/when-yb-remotebuild-fails-to-find-project-in-the-api-show-which-project-it-tried-to-find

Shows up the list of URL used to fetch the project from the build server (API).

This applies over #51

### DIFF
--- a/cli/remote_build.go
+++ b/cli/remote_build.go
@@ -622,6 +622,8 @@ func defineBranch(r *git.Repository, hintBranch string) (string, error) {
 func (p *RemoteCmd) fetchProject(urls []string) (*Project, GitRemote, error) {
 	var empty GitRemote
 	v := url.Values{}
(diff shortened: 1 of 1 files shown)


cli/remote_build.go
```diff
@@ -622,6 +622,8 @@ func defineBranch(r *git.Repository, hintBranch string) (string, error) {
 func (p *RemoteCmd) fetchProject(urls []string) (*Project, GitRemote, error) {
 	var empty GitRemote
 	v := url.Values{}
+	urlsList := fmt.Sprintf("URL's used to do the search: %s", urls)
```
johnewart: Maybe just print this out to avoid multi-line errors? Like "Couldn't find any project using the following URLs...." (also there's no `'` in the plural or URLs)

PR #56, opened 2019-10-25
stable auto release, preview releases as full versions

beholders-eye:
If a tag has '-stable' in it, change channel to stable, and run equinox in our build servers as it would for preview and unstable.

Made preview work as a version, so we can better identify preview versions too, and track them closer to unstable/stable.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-set -x
-
 APP="app_gtQEt1zkGMj"
 PROJECT="yb"
 TOKEN="${RELEASE_TOKEN}"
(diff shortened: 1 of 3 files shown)


release.sh
```diff
@@ -17,20 +17,15 @@ if [ -z "${VERSION}" ]; then
...
 
-if [ "${CHANNEL}" == "preview" ]; then
-  echo "Channel is preview, setting version to timestamp"
-  VERSION="$(date +"%Y%m%d%H%M%S")"
-elif [ "${CHANNEL}" == "stable" ]; then
-  echo "To promote to stable, use the equinox release tool:"
-  echo '$ equinox publish --token $TOKEN --app $APP  --channel stable --release $VERSION'
-  exit 0
+if [ "${STABLE_TAGGED}" == "-stable" ]; then
+    CHANNEL="stable"
```
johnewart: We should probably strip `-stable` from the version string

beholders-eye: Sure

PR #57, opened 2019-10-25
hotfix out of index error

beholders-eye:
Fixes https://app.yourbase.io/yourbase/yb/builds/ac8c9d78-584f-409e-be97-acafadc94ec2


cli/build.go
```diff
@@ -155,7 +155,10 @@ func (b *BuildCmd) Execute(_ context.Context, f *flag.FlagSet, _ ...interface{})
 	for _, envString := range primaryTarget.Environment {
 		parts := strings.Split(envString, "=")
 		key := parts[0]
-		value := parts[1]
+		value := ""
+		if len(parts) > 1 {
```
johnewart: Should we warn that this doesn't look like an environment variable?

PR #59, opened 2019-10-25
aligned with V2 UI

beholders-eye:
That's about right, I think. Now the path scheme is `/organizations/<org slug>/projects/<project label>/builds/<UUID | ID>`. Right?


amyhua:
Yep, that's right

amyhua:
However, for build pages, I'd also like to make `/builds/<uuid>` accessible!

beholders-eye:
> However, for build pages, I'd also like to make `/builds/<uuid>` accessible!

I'll add this when we have it. This new end point can show builds grouped by projects or orgs. Thanks for the review!

PR #60, opened 2019-10-25
pypred may need ()

beholders-eye:
To auto release correctly. Somehow `PyPred` didn't understand `branch IS NOT 'master'` without parenthesis.

### DIFF
--- a/.yourbase.yml
+++ b/.yourbase.yml
@@ -62,11 +62,11 @@ ci:
     - name: test_build
       build_target: default
 
-    - name: pre_release
-      build_target: preview_release
-      when: branch IS NOT 'master' AND tagged IS true
-
     - name: release
       build_target: release
       when: branch IS 'master' AND tagged IS true
+
+    - name: pre_release
+      build_target: preview_release
+      when: (branch IS NOT 'master') AND tagged IS true
 


nictuku:
What was the behavior or error when you tried? How can we help our users not have the same problem? Should yb check-config catch this?

beholders-eye:
It was subtle: https://app.yourbase.io/yourbase/yb/builds/a6277e91-cbea-42a1-bea1-e529b09a813e

Checking the config won't pick it up, the CI should have build the `release` build target instead of the `pre_release` build target. I think it is because of the lack of parenthesis.

nictuku:
I'm not comfortable approving this because you don't seem to have confidence in why this is happening, so I'll defer the review to someone else.

beholders-eye:
https://github.com/armon/pypred/blob/master/README.md says "Parenthesis to disambiguate", so I think it was ambiguous before.

beholders-eye:
> I'm not comfortable approving this because you don't seem to have confidence in why this is happening, so I'll defer the review to someone else.

Right. I'll test it locally.

nictuku:
what was ambiguous? how exactly did it interpret it? etc let's make sure we know what we are doing when committing high-criticality code.

beholders-eye:
> what was ambiguous? how exactly did it interpret it? etc let's make sure we know what we are doing when committing high-criticality code.

The current version uses this predicate: `branch IS NOT 'master' AND tagged IS true`, PyPred picked it in detriment to the other one ('release') because it interpreted `branch IS` as one Boolean expression, `NOT 'master'` as another and `tagged IS true` as yet another one. So to disambiguate, I inserted those parenthesis. And then it should work. But I'm trying to execute it locally.

nictuku:
so are you saying that "foo IS NOT bla" will always be interpreted as to separate boolean expressions?

beholders-eye:
Yes

nictuku:
Are you sure that adding a parenthesis around "(foo IS NOT bla)" help? If I understand this right, it will still be:

(branch IS (NOT 'master'))
(branch IS (false))

Which is still not what we want, or is it?

beholders-eye:
> Are you sure that adding a parenthesis around "(foo IS NOT bla)" help? If I understand this right, it will still be:
> 
> (branch IS (NOT 'master'))
> (branch IS (false))
> 
> Which is still not what we want, or is it?

For a preview release it is what we want.

nictuku:
the reason I'm skeptic about his change is that I assume pypred parses statements from left to right. If that's right, the parenthesis should not make any difference. What did your local tests conclude?

beholders-eye:
Still working on other changes (remote builds, buildagent metadata fixes, and etc.)

beholders-eye:
yourbase/api#54 created to test this out. I didn't understand yet why `branch IS NOT 'master'` doesn't work. And @nictuku, I agree that if pypred parses the predicate from left to right, this should work. I didn't had the time to read it's source, so I decided to close this and open #61

PR #63, opened 2019-10-30
Remove YB CLI container logic to use Narwhal; go 1.13

johnewart:
Easier to make abstractions this way -- also used by the agent. Now this works on macOS when we have side-container dependencies (it uses local port mapping because Docker doesn't expose VM networks)

Update go to 1.13


cli/build.go
```diff
@@ -19,14 +19,14 @@ import (
 	"time"
 
 	"github.com/johnewart/archiver"
+	"github.com/johnewart/narwhal"
```
nictuku: ideally we should add libraries to our own org so folks on the team can help maintain it.

johnewart: Sure, I'll move it in a separate PR :+1:

nictuku:
Assuming you fixed Ridai's crash in the commit above, 👍

PR #64, opened 2019-11-01
Ignore untracked binary files

beholders-eye:
There's no point in adding binary files, there's no way to represent each one contents in a Patch file. But I worked in the `ask-bin-sizelimit` branch, if anyone wants to check how we could handle it (but it's useless anyway).

Fixes [849](https://app.clubhouse.io/yourbaseio/story/849/yb-remotebuild-failed-for-fingo)


cli/remote_build.go
```diff
@@ -415,6 +427,38 @@ func (p *RemoteCmd) traverseChanges(worktree *git.Worktree, saver *WorktreeSave)
 	return
 }
 
+func (p *RemoteCmd) skip(file string, worktree *git.Worktree) bool {
```
johnewart: Can we rename this to shouldSkip and then put the append logic outside this call?

cli/remote_build.go
```diff
@@ -54,6 +55,7 @@ type RemoteCmd struct {
 	committed      bool
 	publicRepo     bool
 	remotes        []GitRemote
+	skipped        []string
```
johnewart: Can we avoid saving this list in the executable struct? We should largely use this structure for the command itself (flags / config primarily)

beholders-eye:
Rebased and test it again using my 7GB sized local repository :wink: . PTAL @johnewart

PR #73, opened 2020-02-25
Updated yb to support creating an image based on a 

jamesnaftel:
container definition+files


cli/package.go
```diff
@@ -21,7 +24,7 @@ type PackageCmd struct {
 func (*PackageCmd) Name() string     { return "package" }
 func (*PackageCmd) Synopsis() string { return "Create a package artifact" }
 func (*PackageCmd) Usage() string {
-	return `package [--target pkg]`
+	return ``
```
beholders-eye: Won't this be a subcommand called "packages"?

PR #76, opened 2020-03-20
remotebuild: Retries reconnecting to the Dispatcher

beholders-eye:
This sends pings to the Dispatcher WebSocket every 5 seconds, if it gets an EOF, it will try reconnecting. It also manages Ctrl+C (SIGTERM) better, and releases `stdout/stderr` after finishing, so the terminal works normally after it. Gives up retrying a connection after 15 times.


cli/remote_build.go
```diff
@@ -886,45 +903,89 @@ func (cmd *RemoteCmd) submitBuild(project *Project, tagMap map[string]string) er
...
 		if err != nil {
 			remoteErrored()
 			return fmt.Errorf("Cannot connect: %v", err)
 		} else {
 
+			// TODO maybe make Dispatcher give better diagnostics somehow
+			go func() {
+				for {
+					select {
+					case <-finnish:
```
johnewart: Typo, should be `finish` (`finished` or `complete` would maybe be best) - this would be people from Finland 😄

cli/remote_build.go
```diff
@@ -946,7 +1007,15 @@ func (cmd *RemoteCmd) submitBuild(project *Project, tagMap map[string]string) er
 					if !buildSuccess {
 						buildSuccess = strings.Count(string(msg), "-- BUILD SUCCEEDED --") > 0
 					}
+					if !buildFailed {
+						buildFailed = strings.Count(string(msg), "-- BUILD FAILED --") > 0
+						if !buildFailed {
+							buildFailed = strings.Count(string(msg), "Patch '' didn't apply cleanly") > 0
```
johnewart: This is okay at the moment but is a little fragile, is there a better way for us to communicate an error? Maybe the dispatcher should have a status channel too in addition to a log stream channel

cli/remote_build.go
```diff
@@ -886,45 +903,89 @@ func (cmd *RemoteCmd) submitBuild(project *Project, tagMap map[string]string) er
...
 		} else {
 
+			// TODO maybe make Dispatcher give better diagnostics somehow
+			go func() {
+				for {
+					select {
+					case <-finnish:
+						return
+					case <-time.After(5 * time.Second):
+						if err := wsutil.WriteClientMessage(conn, ws.OpPing, []byte("hi Dispatcher")); err != nil {
```
johnewart: I'd change this to `remotebuild ping` or something in case we're debugging network traffic

cli/remote_build.go
```diff
@@ -886,45 +903,89 @@ func (cmd *RemoteCmd) submitBuild(project *Project, tagMap map[string]string) er
...
+							if !recentReconnect && reconnectCount < 15 {
+								if remoteProgress != nil {
+									fmt.Println()
+								}
+								log.Tracef("Build not completed, trying to reconnect")
+								conn.Close()
+								close(finnish)
+								recentReconnect = true
+								reconnectCount += 1
+								goto CONN
```
johnewart: goto can be a little dangerous / hard to grok; any way to avoid it here? I can't tell for sure if we can - if it's not easy then this is fine at the moment

cli/remote_build.go
beholders-eye: :smile: :laughing:

cli/remote_build.go
beholders-eye: I was talking about that with @amyhua sometime ago. Dispatcher and/or the API should send a structured/typed message communicating the states transitions. @nictuku made a comment about it too.

cli/remote_build.go
beholders-eye: I don't like it too. Maybe I'll rewrite the whole remotebuild stuff after runtime-context is merged, because it got a little too long. I was thinking of trying to replace `subcommands` with something like `cobra` to manage the CLI subcommands, log output and maybe follow @amyhua's idea on a better CLI tool.

beholders-eye:
@johnewart PTAL, and check my replies to your comments, please.

beholders-eye:
@johnewart PTAL

beholders-eye:
Rebased on the new v0.1/runtime-context-onecommit branch

PR #77, opened 2020-03-20
Fix the download URL to support other than OpenJDK 8

johnewart:
Tested by specifying the latest version of OpenJDK 11 (11.0.6) and it works

### DIFF
--- a/buildpacks/openjdk.go
+++ b/buildpacks/openjdk.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+        "strconv"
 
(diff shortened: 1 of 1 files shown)


buildpacks/openjdk.go
```diff
@@ -13,21 +14,61 @@ import (
...
+	case 9:
+		subVersion = "11"
+	case 10:
+		subVersion = "13.1"
+	case 11:
+		subVersion = "10"
+	case 12:
+		subVersion = "10"
+	case 13:
+		subVersion = "8"
```
beholders-eye: Should probably add:
`case 14: subVersion = "36"`

PR #79, opened 2020-03-30
On HCL-config: Buildpack's sane use of PATH

beholders-eye:
* check if the BuildSpec exists
* Added GetDefaultPath
* Go update
* go get -u
* go mod tidy


.yourbase.yml
```diff
@@ -1,13 +1,12 @@
...
   build:
-    - go:1.13.7
+    - go:1.14.1
     - python:3.6
 
 build_targets:
   - name: default
     host_only: true
     commands:
-      - go get -u github.com/yourbase/narwhal
```
beholders-eye: Fixed our YML and ran `yb-0.0-current run go get -u` followed by `yb-0.0-current run go mod tidy`. Removed import replacements for local versions of Narwhal.

beholders-eye:
This is on top of `hcl-config` which is its turn on top of `runtime-context`. :smile:

PR #90, opened 2020-05-15
Keeps connection with the dispatcher when remote building

beholders-eye:
As https://github.com/yourbase/yb/pull/76#issue-391565652

John asked if we could create a second channel in the Dispatcher to send status to the CLI, I'll work on that.

### DIFF
--- a/cli/remote_build.go
+++ b/cli/remote_build.go
@@ -930,7 +930,7 @@ func (cmd *RemoteCmd) submitBuild(project *Project, tagMap map[string]string) er
 					case <-finish:
 						return
 					case <-time.After(5 * time.Second):
-						if err := wsutil.WriteClientMessage(conn, ws.OpPing, []byte("hi Dispatcher")); err != nil {
+						if err := wsutil.WriteClientMessage(conn, ws.OpPing, []byte("remotebuild ping")); err != nil {
 							log.Errorf("Cannot send ping: %v", err)
 						}
 					}
--- a/cli/remote_build.go
+++ b/cli/remote_build.go
@@ -916,7 +916,7 @@ func (cmd *RemoteCmd) submitBuild(project *Project, tagMap map[string]string) er
 	CONN:
 		conn, _, _, err := ws.DefaultDialer.Dial(context.Background(), url)
 
-		finnish := make(chan struct{})
+		finish := make(chan struct{})
 
 		if err != nil {
 			remoteErrored()
@@ -927,7 +927,7 @@ func (cmd *RemoteCmd) submitBuild(project *Project, tagMap map[string]string) er
 			go func() {
 				for {
 					select {
-					case <-finnish:
+					case <-finish:
 						return
 					case <-time.After(5 * time.Second):
 						if err := wsutil.WriteClientMessage(conn, ws.OpPing, []byte("hi Dispatcher")); err != nil {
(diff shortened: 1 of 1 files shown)


beholders-eye:
Released v0.1.2-preview1 from this branch.

cli/remote_build.go
```diff
@@ -780,12 +782,27 @@ func (cmd *RemoteCmd) savePatch() error {
...
 	}
+	remoteErrored := func() {
+		if remoteProgress != nil {
+			remoteProgress.Fail()
+		}
+	}
+
+	c := make(chan os.Signal)
+	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
+	go func() {
```
zombiezen, line 799: This goroutine worries me for a few reasons:

1. The goroutine will leak past the end of `submitBuild` if an interrupt signal is not sent to the process.
2. The calls to `submitErrored` and `remoteErrored` both race on the assignment to these local variables.
3. It races with indicating success: imagine the case where the build succeeds and then an interrupt signal is sent before the main goroutine is able to report success. It's undefined which one of these would be called first.

I recommend that we start plumbing `Context` throughout the `yb` codebase, which is the standard pattern in Go for dealing with these sorts of issues. See https://talks.golang.org/2014/gotham-context.slide for a brief overview.

cli/remote_build.go
```diff
@@ -13,11 +13,13 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"os/signal"
 	"path"
 	"regexp"
 	goruntime "runtime"
 	"strconv"
 	"strings"
+	"syscall"
```
zombiezen, line 22: nit: It's preferred to use the import `"golang.org/x/sys/unix"` to indicate that this package requires Unix-specific symbols.

cli/remote_build.go
```diff
@@ -888,45 +905,89 @@ func (cmd *RemoteCmd) submitBuild(project *Project, tagMap map[string]string) er
...
+				for {
+					select {
+					case <-finish:
+						return
+					case <-time.After(5 * time.Second):
+						if err := wsutil.WriteClientMessage(conn, ws.OpPing, []byte("remotebuild ping")); err != nil {
+							log.Errorf("Cannot send ping: %v", err)
+						}
+					}
+				}
```
zombiezen, line 937: Using `time.After` leaks a timer when `<-finish` is finally received from. Also, since this is a repeating edge, prefer using [`*time.Ticker`](https://pkg.go.dev/time?tab=doc#Ticker):

```suggestion
				t := time.NewTicker(5 * time.Second)
				defer t.Stop()
				for {
					select {
					case <-finish:
						return
					case <-t.C:
						if err := wsutil.WriteClientMessage(conn, ws.OpPing, []byte("remotebuild ping")); err != nil {
							log.Errorf("Cannot send ping: %v", err)
						}
					}
				}
```

cli/remote_build.go
```diff
@@ -888,45 +905,89 @@ func (cmd *RemoteCmd) submitBuild(project *Project, tagMap map[string]string) er
...
 			buildSuccess := false
+			buildFailed := false
 			buildSetupFinished := false
 			for {
 				msg, control, err := wsutil.ReadServerData(conn)
 				if err != nil {
-					if err != io.EOF {
-						log.Tracef("Unstable connection: %v", err)
-					} else {
+					if err == io.EOF {
```
zombiezen, line 952: In code that targets Go 1.13 or above, prefer using [`errors.Is`](https://pkg.go.dev/errors?tab=doc#Is):

```suggestion
					if errors.Is(err, io.EOF) {
```

cli/remote_build.go
```diff
@@ -888,45 +905,89 @@ func (cmd *RemoteCmd) submitBuild(project *Project, tagMap map[string]string) er
...
+							if !recentReconnect && reconnectCount < 15 {
+								if remoteProgress != nil {
+									fmt.Println()
+								}
+								log.Tracef("Build not completed, trying to reconnect")
+								conn.Close()
+								close(finish)
+								recentReconnect = true
+								reconnectCount += 1
+								goto CONN
```
zombiezen, line 973: This nesting and `goto` makes this logic hard to follow. I recommend trying to factor out some of this into a separate function.

PR #91, opened 2020-05-15
New runtime doesn't support prefixing every command

beholders-eye:
`yb build -exec-prefix <something>` isn't working anymore. This tries to fix it.

### DIFF
--- a/workspace/build_target.go
+++ b/workspace/build_target.go
@@ -3,6 +3,7 @@ package workspace
 import (
 	"fmt"
 	"github.com/yourbase/narwhal"
+	"github.com/yourbase/yb/plumbing"
 	"github.com/yourbase/yb/plumbing/log"
 	"github.com/yourbase/yb/runtime"
 	"os"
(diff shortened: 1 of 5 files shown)


workspace/build_target.go
```diff
@@ -156,10 +156,10 @@ func (bt BuildTarget) Build(runtimeCtx *runtime.Runtime, flags BuildFlags, packa
 
 	// Do this after the containers are up
 	for _, envString := range bt.EnvironmentVariables(runtimeCtx.EnvironmentData()) {
-		parts := strings.Split(envString, "=")
+		parts := strings.SplitN(envString, "=", 2)
```
zombiezen, line 159: Optional: Perhaps this should be extracted out to a function to share with the code in #87?

workspace/build_target.go
```diff
@@ -180,6 +180,10 @@ func (bt BuildTarget) Build(runtimeCtx *runtime.Runtime, flags BuildFlags, packa
 	for _, cmdString := range bt.Commands {
 		var stepError error
 
+		if len(flags.ExecPrefix) > 0 {
+			cmdString = fmt.Sprintf("%s %s", flags.ExecPrefix, cmdString)
```
zombiezen, line 184: nit for clarity:

```suggestion
		if flags.ExecPrefix != "" {
			cmdString = flags.ExecPrefix + " " + cmdString
```

workspace/build_target.go
beholders-eye: Makes sense

beholders-eye:
Addressed @zombiezen PR feedback. Thanks!

PR #92, opened 2020-05-22
Checks if mount's source directory exists

beholders-eye:
If someone removes a bind mount dir of a dependend container, or an exec
container, like `rm -r ${YB_CLI_BUILD_DIR}/containers/<something>`, the
CLI tries to run it, Docker errors saying that this directory doesn't
exist on the host.

This adds a check mounts step when preparing the Exection Environment
for `yb exec`

### DIFF
--- a/workspace/package.go
+++ b/workspace/package.go
@@ -152,14 +152,16 @@ func (p Package) ExecuteToWriter(runtimeCtx *runtime.Runtime, output io.Writer) 
 	return nil
 }
 
-func (p Package) checkMounts(cd narwhal.ContainerDefinition, srcDir string) {
+func (p Package) checkMounts(cd *narwhal.ContainerDefinition, srcDir string) error {
 	for _, mount := range cd.Mounts {
 		parts := strings.Split(mount, ":")
 		if len(parts) == 2 {
 			src := filepath.Join(srcDir, parts[0])
-			MkdirAsNeeded(src)
+			err := MkdirAsNeeded(src)
+			return err
 		}
 	}
+	return nil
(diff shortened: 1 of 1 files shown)


workspace/package.go
```diff
@@ -151,6 +152,16 @@ func (p Package) ExecuteToWriter(runtimeCtx *runtime.Runtime, output io.Writer)
 	return nil
 }
 
+func (p Package) checkMounts(cd narwhal.ContainerDefinition, srcDir string) {
```
zombiezen, line 155: nit: `ContainerDefinition` is a large struct and should generally be passed around by pointer. I know it isn't being passed by value everywhere, but I'm trying to reduce the places where it is.

```suggestion
func (p Package) checkMounts(cd *narwhal.ContainerDefinition, srcDir string) {
```

See https://github.com/golang/go/wiki/CodeReviewComments#receiver-type for a longer discussion.

workspace/package.go
```diff
@@ -151,6 +152,16 @@ func (p Package) ExecuteToWriter(runtimeCtx *runtime.Runtime, output io.Writer)
...
+	for _, mount := range cd.Mounts {
+		parts := strings.Split(mount, ":")
+		if len(parts) == 2 {
+			src := filepath.Join(srcDir, parts[0])
+			MkdirAsNeeded(src)
```
zombiezen, line 160: This error should be checked.

workspace/package.go
beholders-eye: Thanks :+1:

PR #93, opened 2020-06-10
Updated to use Narwhal v0.3.1

beholders-eye:
Still needs more tests


cli/build.go
```diff
@@ -46,7 +46,7 @@ func (b *BuildCmd) SetFlags(f *flag.FlagSet) {
 	f.BoolVar(&b.CleanBuild, "clean", false, "Perform a completely clean build -- don't reuse anything when building")
 }
 
-func (b *BuildCmd) Execute(_ context.Context, f *flag.FlagSet, _ ...interface{}) subcommands.ExitStatus {
+func (b *BuildCmd) Execute(buildCtx context.Context, f *flag.FlagSet, _ ...interface{}) subcommands.ExitStatus {
```
zombiezen, line 49: nit: `Context` parameters are by convention called `ctx`. (here and elsewhere) https://github.com/golang/go/wiki/CodeReviewComments#contexts

runtime/containers.go
```diff
@@ -15,6 +16,7 @@ import (
 )
 
 type ContainerTarget struct {
+	ctx context.Context
```
zombiezen, line 19: Don't store `Context` as a struct field: pass it as a parameter wherever possible. https://github.com/golang/go/wiki/CodeReviewComments#contexts

The main exceptions are:
1. Satisfying an interface that doesn't take in a `Context`
2. Designing code that creates truly asynchronous work like in buildagent.

Passing the `Context` as the first parameter helps the caller arrange for timeouts/cancellation for specific operations, if needed.

runtime/runtime.go
```diff
@@ -188,7 +190,7 @@ type ContainerData struct {
 	serviceCtx *narwhal.ServiceContext
 }
 
-func (c ContainerData) IP(label string) string {
+/*func (c ContainerData) IP(ctx context.Context, label string) string {
```
zombiezen, line 193: If unneeded, delete this function.

runtime/runtime.go
```diff
@@ -55,6 +56,7 @@ type Target interface {
 }
 
 type Process struct {
+	Context     context.Context
```
zombiezen, line 59: Same here: don't store `Context` in a struct field.

workspace/workspace.go
```diff
@@ -22,54 +23,54 @@ type Workspace struct {
 	packages []Package
 }
 
-func (w Workspace) BuildPackage(p Package) error {
+/*func (w Workspace) BuildPackage(p Package) error {
```
zombiezen, line 26: Delete function

runtime/runtime.go
zombiezen: Still applicable.

workspace/workspace.go
zombiezen: Still applicable

PR #94, opened 2020-06-26
Execution runtimes are more reliable

beholders-eye:
This adds a stable ground to running yb run, exec and build
It also does some major clean up.

* Narwhal v0.5.1 was used

[ch815] [ch1592] [ch1594] [ch1593]

### DIFF
--- a/workspace/workspace.go
+++ b/workspace/workspace.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"crypto/sha256"
 	"fmt"
-	"github.com/yourbase/yb/runtime"
 	"io/ioutil"
 	"os"
 	"os/user"
@@ -15,6 +14,7 @@ import (
 
 	. "github.com/yourbase/yb/plumbing"
 	"github.com/yourbase/yb/plumbing/log"
+	"github.com/yourbase/yb/runtime"
 )
 
 type Workspace struct {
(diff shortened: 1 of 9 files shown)


workspace/workspace.go
```diff
@@ -216,15 +226,68 @@ func LoadWorkspace() (Workspace, error) {
...
 
+	var w Workspace
+
 	if PathExists(manifestFile) {
 		log.Debugf("Loading workspace from manifest file: %s", manifestFile)
-		return loadWorkspaceFromPackage(manifestFile, workspacePath)
+		w, err = loadWorkspaceFromPackage(manifestFile, workspacePath)
+	}
+
+	if err == nil && validWorkspaceConfig(w) {
```
zombiezen, line 236: It seems like the less surprising functionality would be to return failure if a configuration file exists but is invalid. WDYT?

workspace/workspace.go
```diff
@@ -216,15 +226,68 @@ func LoadWorkspace() (Workspace, error) {
...
+		return w, nil
+	}
+
+	return Workspace{}, fmt.Errorf("Error resolving workspace at path %s", workspacePath)
+}
+
+// validWorkspaceConfig will check if every expected part of whole Configuration Graph is initialized
+// as expected
+func validWorkspaceConfig(workspace Workspace) (valid bool) {
```
zombiezen, line 254: Consider returning an error (or multiple) from here to help the user in determining what part fails validation.

workspace/workspace.go
beholders-eye: That it could get in the way of the developer/user, but I can come up with a way to do that, will add as a TODO

workspace/workspace.go
beholders-eye: Next PR, got it! :wink:

beholders-eye:
@zombiezen Don't you think this could be approved?

zombiezen:
> @zombiezen Don't you think this could be approved?

Whoops! Forgot to hit the button.

PR #98, opened 2020-06-30
ContainerTarget.WorkDir() was empty

beholders-eye:
This adds a good known default

### DIFF
--- a/runtime/containers.go
+++ b/runtime/containers.go
@@ -279,6 +279,17 @@ func (t *ContainerTarget) WriteFileContents(ctx context.Context, contents string
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


runtime/containers.go
```diff
@@ -279,6 +279,17 @@ func (t *ContainerTarget) WriteFileContents(ctx context.Context, contents string
 	return nil
 }
 
+func (t *ContainerTarget) Init(ctx context.Context, workDir string) error {
```
zombiezen, line 282: Will this method get called as part of an interface? I don't understand how this gets called.

beholders-eye: Man... :sweat_smile:

How is this working then? I'll double check. I thought that Init would be called on initialization

PR #99, opened 2020-06-30
No need to set a useless buffer

beholders-eye:
Uses Narwhal v0.5.2, that fixes how ExecShell passes output and enables
AttachStd{out,err}

[ch1635]


runtime/containers.go
```diff
@@ -82,10 +82,8 @@ func (t *ContainerTarget) ToolsDir(ctx context.Context) string {
 func (t *ContainerTarget) PathExists(ctx context.Context, path string) bool {
 	// Assume we can use stat for now
 	statCmd := fmt.Sprintf("stat %s", path)
-	// Will be ignored, but Docker only succeeds if we pass an output io stream
-	buff := new(bytes.Buffer)
 
-	err := narwhal.ExecShell(ctx, narwhal.DockerClient(), t.Container.Id, statCmd, &narwhal.ExecShellOptions{CombinedOutput: buff})
+	err := narwhal.ExecShell(ctx, narwhal.DockerClient(), t.Container.Id, statCmd, &narwhal.ExecShellOptions{})
```
zombiezen, line 86: optional nit:

```suggestion
	err := narwhal.ExecShell(ctx, narwhal.DockerClient(), t.Container.Id, statCmd, nil)
```

PR #101, opened 2020-07-01
Adds a GoReleaser buildpack

beholders-eye:
A Go project can then add a .goreleaser.yml to setup how it would upload
assets to any SCM provider

Next step: add it to our own release build step

### DIFF
--- a/buildpacks/goreleaser.go
+++ b/buildpacks/goreleaser.go
@@ -1,0 +1,96 @@
+package buildpacks
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+
+	"github.com/yourbase/yb/plumbing/log"
+	"github.com/yourbase/yb/runtime"
+	. "github.com/yourbase/yb/types"
+)
+
+// TODO add yourbase/release-test testing facilities
+
+var GORELEASE_DIST_TEMPLATE_URL = "https://github.com/goreleaser/goreleaser/releases/download/%s/%s"
+
+type GoReleaserBuildTool struct {
+	BuildTool
+	version string
+	spec    BuildToolSpec
+}
+
+func NewGoReleaserBuildTool(toolSpec BuildToolSpec) GoReleaserBuildTool {
+	tool := GoReleaserBuildTool{
+		version: toolSpec.Version,
+		spec:    toolSpec,
+	}
+
+	return tool
+}
+
+func (bt GoReleaserBuildTool) ArchiveFile() string {
+	operatingSystem := bt.spec.InstallTarget.OS()
+	arch := bt.spec.InstallTarget.Architecture()
+	os := "Linux"
+	architecture := "x86_64"
+	ext := "tar.gz"
+
+	if operatingSystem == runtime.Darwin {
+		os = "Darwin"
+	} else if operatingSystem == runtime.Windows {
+		os = "Windows"
+		ext = "zip"
+	}
+
+	// TODO support arm64, armv6
+	if arch != runtime.Amd64 {
+		architecture = "i386"
+	}
+
+	return fmt.Sprintf("goreleaser_%s_%s.%s", os, architecture, ext)
+}
+
+func (bt GoReleaserBuildTool) DonwloadUrl() string {
+	return fmt.Sprintf(GORELEASE_DIST_TEMPLATE_URL, bt.Version(), bt.ArchiveFile())
+}
+
+func (bt GoReleaserBuildTool) Version() string {
+	return bt.version
+}
+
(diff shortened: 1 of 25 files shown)


buildpacks/goreleaser.go
```diff
@@ -0,0 +1,96 @@
...
+	"path/filepath"
+
+	"github.com/yourbase/yb/plumbing/log"
+	"github.com/yourbase/yb/runtime"
+	. "github.com/yourbase/yb/types"
+)
+
+// TODO add yourbase/release-test testing facilities
+
+var GORELEASE_DIST_TEMPLATE_URL = "https://github.com/goreleaser/goreleaser/releases/download/%s/%s"
```
zombiezen, line 15:
```suggestion
const goreleaseDistTemplateURL = "https://github.com/goreleaser/goreleaser/releases/download/%s/%s"
```

buildpacks/goreleaser.go
```diff
@@ -0,0 +1,96 @@
...
+	// TODO support arm64, armv6
+	if arch != runtime.Amd64 {
+		architecture = "i386"
+	}
+
+	return fmt.Sprintf("goreleaser_%s_%s.%s", os, architecture, ext)
+}
+
+func (bt GoReleaserBuildTool) DonwloadUrl() string {
+	return fmt.Sprintf(GORELEASE_DIST_TEMPLATE_URL, bt.Version(), bt.ArchiveFile())
```
zombiezen, line 55: Use `url.PathEscape` to escape these substitutions correctly.

buildpacks/goreleaser.go
```diff
@@ -0,0 +1,96 @@
...
+
+	// TODO support arm64, armv6
+	if arch != runtime.Amd64 {
+		architecture = "i386"
+	}
+
+	return fmt.Sprintf("goreleaser_%s_%s.%s", os, architecture, ext)
+}
+
+func (bt GoReleaserBuildTool) DonwloadUrl() string {
```
zombiezen, line 54: Spelling and [initialism case](https://github.com/golang/go/wiki/CodeReviewComments#initialisms):

```suggestion
func (bt GoReleaserBuildTool) DownloadURL() string {
```

buildpacks/goreleaser.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func (bt GoReleaserBuildTool) Install(ctx context.Context) (error, string) {
+	t := bt.spec.InstallTarget
+
+	installDir := filepath.Join(t.ToolsDir(ctx), "goreleaser", bt.Version())
+
+	if t.PathExists(ctx, installDir) {
+		log.Infof("GoReleaser v%s located in %s!", bt.Version(), installDir)
+	} else {
```
zombiezen, line 69: nit: Rather than enclosing the rest of the function body in an `else`, we can reduce indentation by using an early return.

```suggestion
		log.Infof("GoReleaser v%s located in %s!", bt.Version(), installDir)
		return installDir, nil
	}
```

buildpacks/goreleaser.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (bt GoReleaserBuildTool) DonwloadUrl() string {
+	return fmt.Sprintf(GORELEASE_DIST_TEMPLATE_URL, bt.Version(), bt.ArchiveFile())
+}
+
+func (bt GoReleaserBuildTool) Version() string {
+	return bt.version
+}
+
+func (bt GoReleaserBuildTool) Install(ctx context.Context) (error, string) {
```
zombiezen, line 62: Convention is that `error` is the last return value:

```suggestion
func (bt GoReleaserBuildTool) Install(ctx context.Context) (string, error) {
```

beholders-eye:
> I don't see this type being used anywhere yet. Will that be added in a future PR?

This type of build pack? Well, I'll create a future PR adding it as a dependency for the CLI itself, to release binaries to the GH releases page (it also supports .deb and homebrew, I'll talk to @glacials about it later)

beholders-eye:
PTAL @jamesnaftel

buildpacks/goreleaser.go
```diff
@@ -0,0 +1,99 @@
...
+		os = "Darwin"
+	} else if operatingSystem == runtime.Windows {
+		os = "Windows"
+		ext = "zip"
+	}
+
+	// TODO support arm64, armv6
+	if arch != runtime.Amd64 {
+		architecture = "i386"
+	}
```
jamesnaftel, line 47: This might be more clear as a switch.

PR #103, opened 2020-07-02
Install Mac OS X dependencies

beholders-eye:
Ressurected InstallPlatformDependencies, as a kind of a generic function


beholders-eye:
I'll be honest. Never thought of that. But I think it goes in the same lines of the question about `GOPRIVATE` (and the respective answer). To run Python you'll need that, we're just making it a bit easier, IMO. To run Ruby/Homebrew too. That piece of code was already in YB CLI since the first version.

WDYT @zombiezen ?

beholders-eye:
There's no best answer to this, but this should fix Python installations in Mac OS X. If you like, I can test it better

zombiezen:
I'm with @jamesnaftel on this one. If there are system requirements we need to ensure before using the tool, we should be providing instructions to users or, even better, provide macOS binary packages or `brew` packaging.

beholders-eye:
> I'm with @jamesnaftel on this one. If there are system requirements we need to ensure before using the tool, we should be providing instructions to users or, even better, provide macOS binary packages or `brew` packaging.

For the time being, I'll add instructions to README.md.

@zombiezen what do you mean by macOS binary packages. `yb` is released as Mac OS binary too. Do you meant a `.dmg` or `.pkg`? Did you suggest that we add `/Library/Developer/CommandLineTools/Packages/macOS_SDK_headers_for_macOS_10.14.pkg` to something like a `yb-cli.pkg`?

zombiezen:
It's been a long time since I've done macOS packaging, but IIRC the `.pkg` format permits distributing other `.pkg` files as dependencies and this behaves more nicely with macOS sysadmin things.

README.md
```diff
@@ -161,6 +161,11 @@ You can change yb to other channels by using `-channel`:
 
 Currently, the _inner_ yb that runs inside the build container will use the stable channel by default. To change that, set the environment variable `YB_UPDATE_CHANNEL=unstable` before running yb.
 
+# Mac OS X dependency
```
zombiezen, line 164: Let's add this to the "How to use it" section so the user isn't surprised later.

README.md
```diff
@@ -161,6 +161,11 @@ You can change yb to other channels by using `-channel`:
...
+
+Some build packs (Python, Ruby and Homebrew) needs Mac OS SDK Headers to be installed:
```
zombiezen, line 166: Correct styling is "macOS".

```suggestion
Some build packs (Python, Ruby and Homebrew) needs macOS SDK Headers to be installed:
```

PR #106, opened 2020-07-06
A possibility slipped

beholders-eye:
Better fix the release test

### DIFF
--- a/cli/build.go
+++ b/cli/build.go
@@ -65,14 +65,13 @@ func (b *BuildCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface{
 		if err != nil {
 			log.Errorf("Unable to parse argument: %v", err)
 			return subcommands.ExitFailure
-		} else {
-			if pkgName != "" {
-				pkg, err = ws.PackageByName(pkgName)
-				if err != nil {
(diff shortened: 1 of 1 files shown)


beholders-eye:
Doing this just unbroke staging. The issue is with the package name not the error.

cli/build.go
```diff
@@ -65,12 +65,14 @@ func (b *BuildCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface{
...
 			log.Errorf("Unable to parse argument: %v", err)
 			return subcommands.ExitFailure
-		}
-
-		pkg, err = ws.PackageByName(pkgName)
-		if err != nil {
-			log.Errorf("Unable to find package name %s: %v", pkgName, err)
-			return subcommands.ExitFailure
+		} else {
+			if pkgName != "" {
```
beholders-eye, line 69: pkgName can come out empty, if it does, we need to resolve it to the default value (name of directory, in other words, last component of `$PWD`)

beholders-eye: Yeah, it can come after the whole if-else bellow...

beholders-eye:
PTAL @zombiezen

PR #109, opened 2020-07-08
Download management

beholders-eye:
Checks HTTP status on HEAD
Gives up when the URL gives us garbage
Split download and unexport functionality into another source file


runtime/download.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	// Resolve current homeDir
+	if u, err := user.Current(); err == nil && u.HomeDir != "" {
+		return filepath.Join(u.HomeDir, ".cache", "yourbase")
+	}
+	// Try again
+	if homeDir, exists := os.LookupEnv("HOME"); exists {
+		return filepath.Join(homeDir, ".cache", "yourbase")
+	}
+	return defaultCacheDir
```
zombiezen, line 31: In the spirit of [ch1401](https://app.clubhouse.io/yourbaseio/story/1401), let's use [xdgdir](https://pkg.go.dev/go4.org/xdgdir) to find the cache directory.

```suggestion
	if cacheDir := xdgdir.Cache.Path(); cacheDir != "" {
		return filepath.Join(cacheDir, "yourbase")
	}
	return defaultCacheDir
```

runtime/download.go
```diff
@@ -0,0 +1,120 @@
...
+
+	fileExists := false
+	fileSizeMismatch := false
+
+	// Exists, don't re-download
+	if fi, err := os.Stat(cacheFilename); !os.IsNotExist(err) && fi != nil {
+		fileExists = true
+
+		// try HEAD'ing the URL and comparing to local file
+		resp, err := http.Head(url)
```
zombiezen, line 64: This won't obey the `Context` passed in. Unfortunately, it's a bit more code, but this will need to use [`NewRequestWithContext`](https://pkg.go.dev/net/http?tab=doc#NewRequestWithContext) with [`http.DefaultClient`](https://pkg.go.dev/net/http?tab=doc#DefaultClient).

runtime/download.go
```diff
@@ -0,0 +1,120 @@
...
+		if err == nil {
+			// checks response HTTP status
+			// 404, 500 or others like it
+			if strings.HasPrefix(resp.Status, "40") || strings.HasPrefix(resp.Status, "50") {
```
zombiezen, line 68: While this works, using `resp.StatusCode` is more robust.

```suggestion
			if resp.StatusCode != http.StatusOK {
```

runtime/download.go
```diff
@@ -0,0 +1,120 @@
...
+				return "", fmt.Errorf("%s status %s", url, resp.Status)
+			}
+			if fi.Size() != resp.ContentLength {
+				log.Infof("Re-downloading %s because remote file and local file differ in size", url)
+				fileSizeMismatch = true
+			}
+		} else {
```
zombiezen, line 75: Perform an early return rather than using an `else` clause. https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

runtime/download.go
```diff
@@ -0,0 +1,120 @@
...
+		if err == nil {
+			// checks response HTTP status
+			// 404, 500 or others like it
+			if strings.HasPrefix(resp.Status, "40") || strings.HasPrefix(resp.Status, "50") {
+				return "", fmt.Errorf("%s status %s", url, resp.Status)
+			}
+			if fi.Size() != resp.ContentLength {
+				log.Infof("Re-downloading %s because remote file and local file differ in size", url)
+				fileSizeMismatch = true
+			}
```
zombiezen, line 74: This can be simplified by negating the expression and returning with success. This means the boolean variables can be dropped.

```suggestion
			if fi.Size() == resp.ContentLength {
				log.Infof("Re-using cached version of %s", url)
				return cacheFilename, nil
			}
			log.Infof("Re-downloading %s because remote file and local file differ in size", url)
```

runtime/download.go
```diff
@@ -0,0 +1,120 @@
...
+
+	// Cancellable request
+	req, err := http.NewRequestWithContext(ctx, "GET", url, nil)
+	if err != nil {
+		return err
+	}
+	client := &http.Client{}
+
+	// Get the data
+	resp, err := client.Do(req)
```
zombiezen, line 104: Use `http.DefaultClient`.

```suggestion
	// Get the data
	resp, err := http.DefaultClient.Do(req)
```

runtime/download.go
```diff
@@ -0,0 +1,120 @@
...
+		return err
+	}
+	defer resp.Body.Close()
+
+	// Create the file
+	out, err := os.Create(filepath)
+	if err != nil {
+		return err
+	}
+	defer out.Close()
```
zombiezen, line 115: The `error` from `Close` needs to be checked for correctness. https://www.joeshaw.org/dont-defer-close-on-writable-files/

runtime/download.go
```diff
@@ -0,0 +1,120 @@
...
+func cacheFilenameForURL(url string) (string, error) {
+	reg, err := regexp.Compile("[^a-zA-Z0-9.]+")
+	if err != nil {
+		return "", fmt.Errorf("Can't compile regex: %v", err)
+	}
+	fileName := reg.ReplaceAllString(url, "")
+	return fileName, nil
+}
+
+func downloadFileWithCache(ctx context.Context, url string) (string, error) {
```
zombiezen, line 43: Optional: consider adding unit tests for this function. You can fake an HTTP server with [`httptest.Server`](https://pkg.go.dev/net/http/httptest?tab=doc#Server).

runtime/download.go
beholders-eye: HTTP Statuses 20{1,2,6} etc, would be treated as an error, but this simplistic download manager isn't resuming downloads yet, so moving forward. Thanks!

runtime/download.go
beholders-eye: I've changed GET, but didn't change HEAD. Thanks!

runtime/download.go
beholders-eye: This was the original code, but good suggestion!

runtime/download.go
beholders-eye: sure thing

runtime/download.go
```diff
@@ -0,0 +1,123 @@
...
+
+	if err := os.MkdirAll(cacheDir, 0700); err != nil {
+		return "", fmt.Errorf("creating dir %s: %v", cacheDir, err)
+	}
+
+	cacheFilename := filepath.Join(cacheDir, filename)
+	log.Infof("Downloading %s to cache as %s", url, cacheFilename)
+
+	// Exists, don't re-download
+	if fi, err := os.Stat(cacheFilename); !os.IsNotExist(err) && fi != nil {
```
zombiezen, line 54:
```suggestion
	if fi, err := os.Stat(cacheFilename); err == nil {
```

runtime/download.go
```diff
@@ -0,0 +1,123 @@
...
+		req, err := http.NewRequestWithContext(ctx, "HEAD", url, nil)
+		if err != nil {
+			return "", err
+		}
+
+		// try HEAD'ing the URL and comparing to local file
+		resp, err := http.DefaultClient.Do(req)
+		if err != nil {
+			return "", fmt.Errorf("fetching %s: %v", url, err)
+		}
```
zombiezen, line 65: `resp.Body` needs to be closed, as per https://pkg.go.dev/net/http?tab=doc#Response.Body

```suggestion
		}
		resp.Body.Close()
```

runtime/download_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"net/http"
+	"net/http/httptest"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"go4.org/xdgdir"
+)
+
+func Test_downloadFileWithCache(t *testing.T) {
```
zombiezen, line 15: Optional suggestion for this test: I think the properties most important for this function are

- the file content matches what's on the server
- the GET request only occurs once

The filename returned isn't important to the caller and may change if we decide how and where we store the file.

runtime/download_test.go
```diff
@@ -0,0 +1,46 @@
...
+	contantMock := "Please download me"
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintln(w, contantMock)
+	}))
+
+	defer ts.Close()
+	cachePath := filepath.Join(xdgdir.Cache.Path(), "yourbase")
+
+	cacheFilename, err := downloadFileWithCache(context.Background(), ts.URL)
```
zombiezen, line 24: These files should be removed before the test completes to avoid affecting future runs.

PR #110, opened 2020-07-09
One can use a template string to dynamically add the container set IP Address

beholders-eye:
One can reference a container IP address in the YAML file, using this

```
exec:
  ports:
    - 5000
  dependencies:
    containers:
      db:
        image: postgres:11.0
  environment:
    default:
      - DATABASE_URL=postgres://user:pwd@{{ .Containers.IP "db" }}/basedb

```

### DIFF
--- a/runtime/runtime.go
+++ b/runtime/runtime.go
@@ -172,7 +172,7 @@ func (r *Runtime) Shutdown(ctx context.Context) error {
 }
 
 func (r *Runtime) EnvironmentData() RuntimeEnvironmentData {
-	d := RuntimeEnvironmentData{
+	return RuntimeEnvironmentData{
 		Containers: ContainerData{
 			serviceCtx: r.ContainerServiceContext,
 		},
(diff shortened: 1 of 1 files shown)


runtime/runtime.go
```diff
@@ -196,7 +196,14 @@ type ContainerData struct {
 	serviceCtx *narwhal.ServiceContext
 }
 
+// Environment returns a map of enviroment variables with each container IP adress
+//  it also fill up the ip map, that is used in IP
 func (c ContainerData) Environment(ctx context.Context) map[string]string {
+	// As it's called only once
+	if c.ip == nil {
+		c.ip = make(map[string]string)
```
zombiezen, line 204: I'm not sure I understand the effect of this assignment. From [the spec](https://golang.org/ref/spec#Index_expressions):

> if the map is `nil` or does not contain such an entry, `a[x]` is the zero value for the element type of [the map]

beholders-eye: I was getting a SIGSEV for trying to assign a key+value to a nil map. But that's why this is a draft, I was working a better way to initialize this

zombiezen: That makes sense. In the future, having a new test case in the PR or expanding the description would help me understand the change.

beholders-eye:
PTAL @zombiezen

zombiezen:
Is this still draft or ready for review?

beholders-eye:
Looks better, so I recon it's ready for review

PR #115, opened 2020-07-14
Some tools need an output/home dir for artifacts

beholders-eye:
Some languages/tools/platforms uses a specific directory to store build
artifacts from deps or other resources. Go (before 1.13.x) uses GOPATH,
Ruby has GEM_HOME, Rust has CARGO_HOME, and so forth.

Uses XDG by default instead of manually setting `.yourbase` dirs

Python early return in Setup()

### DIFF
--- a/runtime/containers.go
+++ b/runtime/containers.go
@@ -18,9 +18,10 @@ import (
 
 const (
 	// Linux container defaults
-	containerDefaultToolsDir = "/opt/yb/tools"
-	containerDefaultCacheDir = "/opt/yb/cache"
-	containerDefaultWorkDir  = "/workspace"
+	containerDefaultToolsDir            = "/opt/yb/tools"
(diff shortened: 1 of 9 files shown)


runtime/containers.go
```diff
@@ -80,6 +81,13 @@ func (t *ContainerTarget) ToolsDir(ctx context.Context) string {
 	return containerDefaultToolsDir
 }
 
+func (t *ContainerTarget) ToolOutputSharedDir(ctx context.Context) string {
+	err := narwhal.MkdirAll(ctx, narwhal.DockerClient(), t.Container.Id, containerDefaultToolOutputSharedDir)
+	if err != nil {
+		return ""
+	}
```
glacials, line 88: Do we need to log this error, or is it an expected case?

beholders-eye: Good catch, thanks

PR #118, opened 2020-07-15
LoadWorkspace should stay in workspace

beholders-eye:
Code move, some more `fmt.Errorf` fixes. The idea is to debug and fix a
project that would have nested `.yourbase.yml` files

Also unexports workspace loading symbols

### DIFF
--- a/workspace/workspace.go
+++ b/workspace/workspace.go
@@ -15,6 +15,7 @@ import (
 	. "github.com/yourbase/yb/plumbing"
 	"github.com/yourbase/yb/plumbing/log"
 	"github.com/yourbase/yb/runtime"
+	. "github.com/yourbase/yb/types"
 )
 
 type Workspace struct {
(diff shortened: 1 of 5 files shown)


workspace/workspace.go
```diff
@@ -216,11 +217,15 @@ func loadWorkspaceFromConfigYaml(configFile string, path string) (Workspace, err
 }
 
 func LoadWorkspace() (Workspace, error) {
+	return loadWorkspace()
+}
```
glacials, line 221: What's the purpose of having both an exported and a non-exported version of this function?

workspace/workspace.go
```diff
@@ -298,3 +303,59 @@ func validWorkspaceConfig(workspace Workspace) error {
...
+
+	for {
+		file_path := filepath.Join(wd, filename)
+		if _, err := os.Stat(file_path); err == nil {
+			return wd, nil
+		}
+
+		wd = filepath.Dir(wd)
+
+		if strings.HasSuffix(wd, "/") {
```
glacials, line 353: This tripped me up a bit until I researched it, can we add a comment that only root will ever have a trailing `/`?

workspace/workspace.go
```diff
@@ -298,3 +303,59 @@ func validWorkspaceConfig(workspace Workspace) error {
...
+			return "", fmt.Errorf("searching %s, ended up at the root", filename)
+		}
```
glacials, line 355: Not sure it matters a ton, but this will fail to find the file if it's in `/` directly.

workspace/workspace.go
beholders-eye: There's still other places where the exported version is used. I'd like to change this, for the time being, I'll keep it like that.

workspace/workspace.go
beholders-eye: This code was the original, written by John. I never realized that this will fail.

PR #119, opened 2020-07-17
Clean up plumbing.IsBinary

zombiezen:


plumbing/util.go
```diff
@@ -179,36 +177,27 @@ func CloneRepository(remote GitRemote, inMem bool, basePath string) (rep *git.Re
...
-			break
-		}
-
-		b, err := reader.ReadByte()
-		if err == io.EOF {
-			break
-		}
-		if err != nil {
-			return false, err
+	buf := make([]byte, 8000)
```
beholders-eye, line 188: Why do you tend to replace constants with literals?

plumbing/util.go
```diff
@@ -179,36 +177,27 @@ func CloneRepository(remote GitRemote, inMem bool, basePath string) (rep *git.Re
...
+	n, err := io.ReadFull(r, buf)
+	if err != nil {
+		if errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) {
```
beholders-eye, line 191: That will surely improve performance, thx!

plumbing/util.go
zombiezen: I try to scope constants down to as close to where they are needed as possible. Because this is used once in a single function and the number is easy to understand in context, I inlined. If the number had been needed multiple times inside the same function, I would have created a named `const` inside the function scope.

PR #120, opened 2020-07-17
Fix for /tmp permissions issue

beholders-eye:
Also adds more error checks

Tries to fix other permission issues when the container run creates
artifacts, by running inside the container as the current user

Rust's CARGO_HOME/RUSTUP_HOME should be shown to the user

### DIFF
--- a/.yourbase.yml
+++ b/.yourbase.yml
@@ -10,6 +10,7 @@ build_targets:
       - bash -c "go build -ldflags \"-X 'main.date=$(date)'\""
 
   - name: integration
+    host_only: true
     commands:
       - bash -c "cd integration-tests; yb build -no-container"
       - bash -c "cd integration-tests; yb build -no-container python"
--- a/buildpacks/android_sdk.go
+++ b/buildpacks/android_sdk.go
@@ -82,7 +82,7 @@ func (bt AndroidBuildTool) writeAgreements(ctx context.Context, androidDir strin
 
 	licensesDir := filepath.Join(androidDir, "licenses")
 	t := bt.spec.InstallTarget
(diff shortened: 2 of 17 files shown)


beholders-eye:
Just to add some more context.

When a YAML has `apt-get ...` apt libraries tries to create `/tmp/apt-*` index files and fails, because the CLI is injecting its tools into `/tmp`, I changed it to use `/var/tmp/yb-inject` instead. So Narwhal's `archiveFile` function won't try to change `/tmp` permissions bits to `0750` instead of `1777`

zombiezen:
Is this ready for review? It's still in draft.

beholders-eye:
It was, then I found some issues testing it a bit more. I'm addressing those.

beholders-eye:
The last commit, added yesterday, has a lot more than just fixes for `/tmp`. It adds fixed up commands to prepare a build container to run as normal user, using the same UID/GID that of the current local user (still fixed to Linux only). PTAL @jamesnaftel, @zombiezen or @glacials

beholders-eye:
Dependent on `narwhal-v0.5.3`. PTAL, anyone :wink: :waffle:

.yourbase.yml
```diff
@@ -10,6 +10,7 @@ build_targets:
       - bash -c "go build -ldflags \"-X 'main.date=$(date)'\""
 
   - name: integration
+    host_only: true
```
beholders-eye, line 13: I'll work on a way to inject `yb` itself in the container, so this would work

zombiezen:
There is a lot going on in this PR. Can you break this down into smaller PRs for the individual fixes? That would help me review.

PR #126, opened 2020-07-28
Integration tests for main platforms/languages

beholders-eye:
NOTE: This is for the stable version (v0.0.x series) of YB

Adds submodules for some projects, yourbasefyed them to exercise
build packs for:
 * Python (ansible needs some more work, but getting there) (We'll probably need something like Travis's "matrix" construct)
 * Go
 * Java (Gradle and Maven)
 and
 * Ruby
 * Flutter (Gradle too)

### DIFF
--- a/integration-tests/flutter/.yourbase.pin_code_test_field.yml
+++ b/integration-tests/flutter/.yourbase.pin_code_test_field.yml
@@ -1,0 +1,9 @@
+dependencies:
+  build:
+    - gradle:6.5.1
+    - flutter:1.17.5
+
+build_targets:
+  - name: default
+    commands:
+      - flutter pacakges get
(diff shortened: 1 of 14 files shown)


integration-tests/flutter/.yourbase.pin_code_test_field.yml
```diff
@@ -0,0 +1,9 @@
+dependencies:
+  build:
+    - gradle:6.5.1
+    - flutter:1.17.5
+
+build_targets:
+  - name: default
+    commands:
+      - flutter pacakges get
```
zombiezen, line 9:
```suggestion
      - flutter packages get
```

integration-tests/go/.yourbase.gg.yml
```diff
@@ -0,0 +1,8 @@
+dependencies:
+  build:
+    - go:1.14.4
+
+build_targets:
+  - name: default
+    commands:
+      - go test -mod=readonly ./...
```
zombiezen, line 8: A lot of the gg tests depend on having Git installed. You may want to add an `apt-get install git` line before this so that those don't get skipped.

PR #129, opened 2020-07-28
Enables per-build_target dependencies (stable)

beholders-eye:
Add support for overriding and merging build packs defined inside a
`build_target` as in:

```
dependencies:
  build:
    - go:1.13.2

build_targets:
  - name: default
    commands:
      - go build

  - name: next
    dependencies:
      build:
        - go:1.14.6
    commands:
      - go build
```

* Integration test for flutter changed to test it

PR #132, opened 2020-07-29
Enable per-build_target dependencies (stable)

beholders-eye:
Effectively introduces new build deps
 
A build dependency defined inside a build target that isn't defined
globally should be installed for use inside the build. So something
like:

```
dependencies:
  build:
    - go:1.13.4

build_targets:
  - name: default
    commands:
      - go build

  - name: deploy
    dependencies:
      build:
        - python:3.8
        - go:1.14.6
```

Will override the Go version and add Python 3.8.x as a build tool available
for the "deploy" build target.

NOTE: A build target doesn't use runtime dependencies, so there's no
"run:" section inside a build target's dependencies section.

Short-term fix for:
"As Robson I'd like to run different flutter versionson
different builds [ch1868]"

Rebase of #129


packages/package.go
```diff
@@ -15,9 +15,10 @@ import (
 )
 
 type Package struct {
-	Name     string
-	Path     string
-	Manifest BuildManifest
+	Name         string
+	Path         string
+	Manifest     BuildManifest
+	BuildTargets []BuildTarget
```
zombiezen, line 21: IIUC this is effectively taking a previously immutable configuration AST and giving it a mutable parameter to have a side-channel to pass the information. This creates a hidden contract with the caller that this field must be set. It also defers potential problems until build time.

A more robust way to handle this would be to have `LoadPackage` perform the merging so that each individual build target has a full list of its dependencies. Ideally, we'd remove the `BuildManifest.Dependencies` field to avoid accidental misuse, but it's currently entangled with parsing so we can leave that for a future PR.

types/dependencies.go
```diff
@@ -0,0 +1,58 @@
...
+		depMap[tool] = version
+	}
+	// Doing this before makes locally defined deps to override the global definition
+	for _, tgt := range targetList {
+		for _, dep := range tgt.Dependencies.Build {
+			tool, version, err := splitToolName(dep)
+			if err != nil {
+				return nil, err
+			}
+			depMap[tool] = version
```
zombiezen, line 48: Actually, there's a somewhat subtle issue with this logic: we probably want to return an error if we try to build two targets that have different versions for the same build pack dependency. As written, this would use the last target's version (which could cause unexpected behavior).

zombiezen:
BTW, in the future, you don't have to close a PR to perform a rebase or change the base branch. Rebasing can work by using `git push --force-with-lease origin HEAD:mytopicbranch`. You can use the "Edit" button at the top of the GitHub PR UI to change the base branch to merge into.

beholders-eye:
> BTW, in the future, you don't have to close a PR to perform a rebase or change the base branch. Rebasing can work by using `git push --force-with-lease origin HEAD:mytopicbranch`. You can use the "Edit" button at the top of the GitHub PR UI to change the base branch to merge into.
> 
> Closing a PR drops a bunch of state for your reviewer like which files have been viewed and any draft comments.

I never noticed that clicking on "edit" would render it possible to change the base branch :astonished: 

Will do the next time.

types/dependencies.go
beholders-eye: Gotcha

packages/package.go
beholders-eye: Thanks so much for this, Ross. I'll tackle it.

beholders-eye: I think we can do both: `LoadPackage` performs the merge, but a user can keep describing global build dependencies in `BuildManifest.Dependencies`. Also please remember that `BuildManifest.Dependencies.Run` exists and is used by `yb run` and `yb exec`. So it shouldn't go, IMO.

zombiezen: Oops, yeah `Dependencies.Run` should stick around. Overall, the point I was trying to make is that the on-disk representation of the manifest shouldn't be coupled to the loaded representation of the data.

beholders-eye:
@zombiezen or @jamesnaftel, please take another look.

zombiezen:
I still think my previous two comments need to be addressed before merging this in. Let me know if I can help: maybe we could pair-program this one.

PR #133, opened 2020-07-30
Stable releases won't use `-stable` suffix

beholders-eye:
Revamped of the `release.sh` to support:
1. Locally testing it
1. Staging/Preview releases will no-op
1. Best practices bash scripting, thanks @zombiezen
1. Commit Hash info when `yb version`
1. Release candidates, e.g.: "rc1" or "preview1" for pre-releases
1. Officially state that we only keep two channels: "stable" and
   "preview" (pre-release)

NOTE: This is for the stable release after #132 aproval
Fixes [ch1773]


release.sh
```diff
@@ -1,67 +1,59 @@
...
+fi
 
-tmpkeyfile="$(mktemp)"
-trap "cleanup" INT TERM EXIT
+# Commit info
+COMMIT="${YB_GIT_COMMIT:-}"
+if [ -z "${COMMIT}" ]; then
+    # If git is installed
+    if hash git; then
+        COMMIT="$(git rev-list -n1 HEAD)"
```
zombiezen, line 43: Slightly more compact way to write:

```suggestion
        COMMIT="$(git rev-parse HEAD)"
```

release.sh
```diff
@@ -1,67 +1,59 @@
...
-    rv=$?
-    rm -rf "$tmpkeyfile"
-    exit $rv
-}
-
-set -x
+local_test_release="${TEST_RELEASE:-}"
+
+# YourBase S3 artifacts Release
+# TODO migrate to CATS
```
zombiezen, line 27: This now has its [own story](https://app.clubhouse.io/yourbaseio/story/1940/automate-yb-release-uploads-to-cats):

```suggestion
# TODO(ch1940): migrate to CATS
```

PR #139, opened 2020-08-06
CATS Release!

beholders-eye:
Sends yb releases to the new CATS bucket so, it can be deployed through
CATS ;)

### DIFF
--- a/package.sh
+++ b/package.sh
@@ -29,7 +29,7 @@ do
     out="$( pwd )"
     ( cd "$PKG_DIR" && zip -r "$out/$BUNDLE.zip" "$BUNDLE" )
 
-    # TODO(ch1844): Remove after being sure no build server needs this anymore
+    # TODO(ch2141): Remove after being sure no build server needs this anymore
     mv "$PKG_DIR/$BUNDLE/yb" "$PKG_DIR/$BUNDLE/yb-${os}-${arch}"
     tar -C "$PKG_DIR/$BUNDLE" -zcvf "$out/yb-${os}-${arch}-${VERSION}.tgz" "yb-${os}-${arch}"
 
--- a/package.sh
+++ b/package.sh
@@ -29,7 +29,7 @@ do
     out="$( pwd )"
     ( cd "$PKG_DIR" && zip -r "$out/$BUNDLE.zip" "$BUNDLE" )
 
-    # TODO(ch1844): Remove once CATS is fully rolled out.
+    # TODO(ch1844): Remove after being sure no build server needs this anymore
     mv "$PKG_DIR/$BUNDLE/yb" "$PKG_DIR/$BUNDLE/yb-${os}-${arch}"
     tar -C "$PKG_DIR/$BUNDLE" -zcvf "$out/yb-${os}-${arch}-${VERSION}.tgz" "yb-${os}-${arch}"
 
--- a/package.sh
+++ b/package.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+set -euo pipefail
+
+. variables.sh
+
+PKG_DIR="$(mktemp -d)"
+cleanup() {
+  rm -rf "$PKG_DIR"
+}
+trap cleanup EXIT
+
+
+OSLIST=( linux darwin )
+ARCHLIST=( amd64 )
+
+for os in "${OSLIST[@]}"
+do
+  for arch in "${ARCHLIST[@]}"
+  do
+    BUNDLE="yb_${VERSION}_${os}_${arch}"
+    echo "Building yb package $VERSION in $PKG_DIR/$BUNDLE..."
+
+    GOOS=${os} GOARCH=${arch} go build -ldflags \
+    "-X 'main.version=$VERSION' -X 'main.date=$(date -u '+%F-%T')' -X 'main.channel=$CHANNEL'${BUILD_COMMIT_INFO} -s -w" \
+    -o "$PKG_DIR/$BUNDLE/yb"
+
+    # CATS
+    out="$( pwd )"
+    ( cd "$PKG_DIR" && zip -r "$out/$BUNDLE.zip" "$BUNDLE" )
+
+    # TODO(ch1844): Remove once CATS is fully rolled out.
+    mv "$PKG_DIR/$BUNDLE/yb" "$PKG_DIR/$BUNDLE/yb-${os}-${arch}"
+    tar -C "$PKG_DIR/$BUNDLE" -zcvf "$out/yb-${os}-${arch}-${VERSION}.tgz" yb-${os}-${arch}
+
+  done
+done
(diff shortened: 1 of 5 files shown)


package.sh
```diff
@@ -0,0 +1,38 @@
...
+    "-X 'main.version=$VERSION' -X 'main.date=$(date -u '+%F-%T')' -X 'main.channel=$CHANNEL'${BUILD_COMMIT_INFO} -s -w" \
+    -o "$PKG_DIR/$BUNDLE/yb"
+
+    # CATS
+    out="$( pwd )"
+    ( cd "$PKG_DIR" && zip -r "$out/$BUNDLE.zip" "$BUNDLE" )
+
+    # TODO(ch1844): Remove once CATS is fully rolled out.
+    mv "$PKG_DIR/$BUNDLE/yb" "$PKG_DIR/$BUNDLE/yb-${os}-${arch}"
+    tar -C "$PKG_DIR/$BUNDLE" -zcvf "$out/yb-${os}-${arch}-${VERSION}.tgz" yb-${os}-${arch}
```
zombiezen, line 34: nit: quotes here:

```suggestion
    tar -C "$PKG_DIR/$BUNDLE" -zcvf "$out/yb-${os}-${arch}-${VERSION}.tgz" "yb-${os}-${arch}"
```

release.sh
```diff
@@ -35,76 +17,31 @@ if $aws_disabled; then
...
-        fi
-
-        aws s3 cp $i s3://yourbase-artifacts/yb/${VERSION}/
-      else
-        echo "Local test, would run:"
-        echo aws s3 ls s3://yourbase-artifacts/yb/${VERSION}/$i
-        echo aws s3 cp $i s3://yourbase-artifacts/yb/${VERSION}/
-      fi
+    for r in yb-*-*-${VERSION}.*; do
+        for bucket in "yourbase-artifacts/yb/${VERSION}/ yourbase-cats-bundles/"; do
```
zombiezen, line 24: Missing quotes (will only loop once, this is a logic error):

```suggestion
        for bucket in "yourbase-artifacts/yb/${VERSION}/" "yourbase-cats-bundles/"; do
```

release.sh
```diff
@@ -35,76 +17,31 @@ if $aws_disabled; then
...
-            exit 1
-        fi
-
-        aws s3 cp $i s3://yourbase-artifacts/yb/${VERSION}/
-      else
-        echo "Local test, would run:"
-        echo aws s3 ls s3://yourbase-artifacts/yb/${VERSION}/$i
-        echo aws s3 cp $i s3://yourbase-artifacts/yb/${VERSION}/
-      fi
+    for r in yb-*-*-${VERSION}.*; do
```
zombiezen, line 23: CATS exclusively uses zip files: we shouldn't be uploading the tarballs as well.

release.sh
```diff
@@ -35,76 +17,31 @@ if $aws_disabled; then
...
+        for bucket in "yourbase-artifacts/yb/${VERSION}/ yourbase-cats-bundles/"; do
+            if [ -z "${local_test_release}" ]; then
+                aws s3 cp $i s3://${bucket}
```
zombiezen, line 26: Undefined variable:

```suggestion
                aws s3 cp "$r" "s3://$bucket"
```

variables.sh
```diff
@@ -0,0 +1,38 @@
...
+  echo "Extracting version from tag ref ${YB_GIT_BRANCH:-}" 1>&2
+  VERSION="$(echo "${YB_GIT_BRANCH:-}" | sed -e 's|refs/tags/||g')"
+fi
+
+if [ -z "$VERSION" ]; then
+  echo "No version provided, won't release" 1>&2
+  exit 1
+fi
+
+if $(echo $VERSION | grep -vqo '^v'); then
```
zombiezen, line 13: Using `$()` will execute the stdout, which I don't think is what we want here.

```suggestion
if echo "$VERSION" | grep -vqo '^v'; then
```

variables.sh
```diff
@@ -0,0 +1,38 @@
...
+    echo "Doesn't start with a \"v\" when it should, not releasing"
+    exit 1
+fi
+
+if $(echo $VERSION | grep -qo '\-[a-z]\+[0-9]*$'); then
```
zombiezen, line 18: Same here:

```suggestion
if echo "$VERSION" | grep -qo '\-[a-z]\+[0-9]*$'; then
```

package.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+set -euo pipefail
+
+. variables.sh
```
zombiezen, line 5: nit: expand out `source` for clarity.

```suggestion
source variables.sh
```

release.sh
```diff
@@ -2,29 +2,11 @@
...
-fi
-
-if $(echo $VERSION | grep -qo '\-[a-z]\+[0-9]*$'); then
-    # Release candidate release
-    CHANNEL="preview"
-else
-    # Stable releases should not have a suffix
-    CHANNEL="stable"
-fi
+. variables.sh
```
zombiezen, line 5: nit: expand out `source` for clarity.

```suggestion
source variables.sh
```

variables.sh
beholders-eye: Yes, that's what we want, `grep -q` doesn't print anything, it only returns a exit code

zombiezen: I agree we want to run the grep, but `if` decides based on exit code. For example, this conditional will evaluate the _output_ of the echo as a command:

```bash
if $( echo /bin/false ); then
  echo "This won't be run"
fi
```

Whereas this uses the exit code of `echo` and thus will be run:

```bash
if echo /bin/false; then
  echo "This will be run"
fi
```

release.sh
beholders-eye: Strange, I thought I fixed this

variables.sh
beholders-eye: You're right, sorry for the stubbornness :sweat_smile:

beholders-eye:
PTAL @zombiezen

package.sh
```diff
@@ -0,0 +1,37 @@
...
+
+    GOOS="$os" GOARCH="$arch" go build -ldflags \
+    "-X 'main.version=$VERSION' -X 'main.date=$(date -u '+%F-%T')' -X 'main.channel=$CHANNEL'${BUILD_COMMIT_INFO} -s -w" \
+    -o "$PKG_DIR/$BUNDLE/yb"
+
+    # CATS
+    out="$( pwd )"
+    ( cd "$PKG_DIR" && zip -r "$out/$BUNDLE.zip" "$BUNDLE" )
+
+    # TODO(ch1844): Remove after being sure no build server needs this anymore
```
zombiezen, line 32: Let's link to a new story for this one.

PR #144, opened 2020-08-09
plumbing: fix dropped error

alrs:
This fixes a dropped error in the `plumbing` package by logging it (as in surrounding code) and returning an empty string.


beholders-eye:
Nice catch!

PR #145, opened 2020-08-11
Uses `git` to fetch and compare remote branches

beholders-eye:
Instead of using go-git.v4. This is now the default, an user can call
`yb remotebuild -go-git` to override this and switch back to go-git.


cli/best_ancestor.go
```diff
@@ -118,34 +103,92 @@ func fastFindAncestor(r *git.Repository) (h plumbing.Hash, c int, branchName str
...
+			err = fmt.Errorf("git fetch: %w", err)
+			return
+		}
+
+		if buf.Len() == 0 {
+			err = fmt.Errorf("git fetch: empty response")
+			return
+		}
+		branchFetched := strings.Count(buf.String(), "\n")
+		log.Infof("Feched %d branches", branchFetched)
```
zombiezen, line 155:
```suggestion
		log.Infof("Fetched %d branches", branchFetched)
```

cli/best_ancestor.go
```diff
@@ -118,34 +103,92 @@ func fastFindAncestor(r *git.Repository) (h plumbing.Hash, c int, branchName str
...
+			if rerr != nil {
+				if errors.Is(rerr, io.EOF) {
+					err = fmt.Errorf("reached %w", rerr)
+				} else {
+					err = rerr
+				}
+				return
+			}
+			line = strings.Trim(line, " \n")
+			bName := strings.ReplaceAll(line, "origin/", "")
```
zombiezen, line 181: Depending on a particular remote name can be quite brittle. Is there a way we can get around this?

cli/best_ancestor.go
```diff
@@ -118,34 +103,92 @@ func fastFindAncestor(r *git.Repository) (h plumbing.Hash, c int, branchName str
...
+				if strings.HasSuffix(rem.Name().Short(), reference.Name().Short()) {
+					remoteBranch = rem
+					return storer.ErrStop
+				}
+			}
+			return nil
+		})
+	} else {
+		const (
+			gitFetchCmd  = " fetch --all"
```
zombiezen, line 138: If you're just looking to list the remote branches, you can use [`git ls-remote`](https://git-scm.com/docs/git-ls-remote.html) without performing a fetch.

cli/best_ancestor.go
```diff
@@ -118,34 +103,92 @@ func fastFindAncestor(r *git.Repository) (h plumbing.Hash, c int, branchName str
...
 
-func findRemoteBranch(reference *plumbing.Reference, r *git.Repository) (remoteBranch *plumbing.Reference, err error) {
+func (p *RemoteCmd) findRemoteBranch(ctx context.Context, reference *plumbing.Reference, r *git.Repository) (remoteBranch *plumbing.Reference, err error) {
 	branchIter, err := r.References()
 	if err != nil {
 		return
 	}
 	// If branch is main, just point to it directly
-	remoteDefault := "refs/remote/origin/main"
+	remoteDefault := "refs/remotes/origin/main"
```
zombiezen, line 112: I know this is somewhat unrelated to this PR, but it seems like this should be using the remote's `HEAD` rather than depending on certain branch names.

.yourbase.yml
```diff
@@ -1,7 +1,6 @@
 dependencies:
```
zombiezen, line 1: These changes seem unrelated. Revert file?

variables.sh
```diff
@@ -0,0 +1,38 @@
+# This should be included in both scripts, not invoked directly
```
zombiezen, line 1: Unrelated; revert file.

cli/best_ancestor.go
beholders-eye: We need to perform a fetch, because we don't know for sure if the local clone was a shallow one or not. (But, if he/she created a new topic branch, most probably the local repo, at `$(pwd)/.git` has enough info to find it... got your point, I didn't thought so previously). This could simplify remote build logic in other places too.

variables.sh
beholders-eye: Yeah, I took the opportunity to remote build `local_test_release` and check if per-build target was indeed working. And it actually didn't, and I'm still trying to understand why:

https://app.staging.yourbase.io/yourbase/yb/builds/64653c0d-58e8-4b8d-ac41-8ce7856c1c54

beholders-eye:
PTAL @zombiezen

PR #147, opened 2020-08-13
Appropriate way of finding a running container IP address

beholders-eye:
This will effectively replace "{{ .Containers.IP "xxx" }}"  with the docker network IP for a running container labeled "xxx"

### DIFF
--- a/runtime/runtime.go
+++ b/runtime/runtime.go
@@ -226,11 +226,11 @@ func (c ContainerData) IP(label string) string {
 				break
 			}
 		}
-		container, err := c.serviceCtx.FindContainer(context.Background(), &containerDef)
+		container, err := c.serviceCtx.FindContainer(context.TODO(), &containerDef)
 		if err != nil {
 			return ""
 		}
-		if ipv4, err := narwhal.IPv4Address(context.Background(), c.serviceCtx.DockerClient, container.Id); err == nil {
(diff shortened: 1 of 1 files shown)


runtime/runtime.go
```diff
@@ -214,11 +218,21 @@ func (c ContainerData) Environment(ctx context.Context) map[string]string {
...
+			if label == c.Label {
+				containerDef = c
+				break
 			}
 		}
+		container, err := c.serviceCtx.FindContainer(context.Background(), &containerDef)
+		if err != nil {
+			return ""
+		}
+		if ipv4, err := narwhal.IPv4Address(context.Background(), c.serviceCtx.DockerClient, container.Id); err == nil {
```
zombiezen, line 233: nit: This seems more like the `Context` isn't being plumbed, so let's make that explicit:

```suggestion
		container, err := c.serviceCtx.FindContainer(context.TODO(), &containerDef)
		if err != nil {
			return ""
		}
		if ipv4, err := narwhal.IPv4Address(context.TODO(), c.serviceCtx.DockerClient, container.Id); err == nil {
```

beholders-eye: FYI, I can't think of any way to plumb in a `Context` here, as this function is called when interpolating `{{ .Containers.IP "xxx" }}`. If I add this, it'd stop working. Or am I missing something here?

zombiezen: You are correct, but I'm saying that `context.TODO` is the correct function to use when we really ought to be plumbing a `Context`, but can't think of a good way to do it right away.

PR #148, opened 2020-08-13
Bumps conda tool version

beholders-eye:
Helps reduce time spent on Python projects


beholders-eye:
Didn't work :-1: :disappointed:

PR #152, opened 2020-08-18
Still needs to point to main

beholders-eye:


zombiezen:
I actually don't think this is the right fix: this should be reading from remote `HEAD`, rather than special-casing the names.

PR #162, opened 2020-08-26
Update channel info

beholders-eye:
Updates information about which channels we're actively working on.


nictuku:
I'm not a good person to review this right now. Assigning to Ross.

PR #167, opened 2020-09-03
Supports setting local environment vars

beholders-eye:
Uses `.env` from current work dir to override or set new environment variables


workspace/build_target_test.go
```diff
@@ -0,0 +1,114 @@
...
+		Name        string
+		Tools       []string
+		Commands    []string
+		Environment []string
+	}
+	type args struct {
+		data runtime.RuntimeEnvironmentData
+	}
+	// Sets up dotenv
+	cwd, err := filepath.Abs(".")
```
zombiezen, line 25: nit: this can made more direct with `os.Getwd`:

```suggestion
	cwd, err := os.Getwd()
```

workspace/build_target_test.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		t.Fatalf("Unable to find current work dir: %v", err)
+	}
+	testPath := filepath.Join(cwd, "TESTING")
+	err = os.Mkdir(testPath, 0755)
+	if err != nil {
+		t.Fatalf("Unable to make test dir: %v", err)
+	}
+	dotEnvFilePath := filepath.Join(testPath, ".env")
+	err = os.Chdir(testPath)
```
zombiezen, line 35: Rather than calling `os.Chdir`, is there a way we can change the system-under-test to permit passing in a file path? IIUC the non-test callers of the function can just always pass in `".env"` because that will use the file in the current directory.

workspace/build_target_test.go
```diff
@@ -0,0 +1,114 @@
...
+				Name:        tt.fields.Name,
+				Tools:       tt.fields.Tools,
+				Commands:    tt.fields.Commands,
+				Environment: tt.fields.Environment,
+			}
+			sort.Slice(tt.want, func(i, j int) bool { return tt.want[i] < tt.want[j] })
+			got := bt.EnvironmentVariables(tt.args.data)
+			sort.Slice(got, func(i, j int) bool { return got[i] < got[j] })
+
+			if !reflect.DeepEqual(got, tt.want) {
```
zombiezen, line 109: Rather than `reflect.DeepEqual`, prefer using [`cmp.Diff`](https://pkg.go.dev/github.com/google/go-cmp/cmp?tab=doc#Diff). It's much easier to view test failures and can handle the slice sorting for you.

workspace/build_target_test.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	type args struct {
+		data runtime.RuntimeEnvironmentData
+	}
+	// Sets up dotenv
+	cwd, err := filepath.Abs(".")
+	if err != nil {
+		t.Fatalf("Unable to find current work dir: %v", err)
+	}
+	testPath := filepath.Join(cwd, "TESTING")
```
zombiezen, line 29: With Go 1.15, you can use [`t.TempDir()`](https://pkg.go.dev/testing?tab=doc#T.TempDir) to procure a directory that gets cleaned up automatically.

workspace/build_manifest.go
```diff
@@ -59,41 +58,53 @@ type ExecPhase struct {
 
 func (e *ExecPhase) EnvironmentVariables(ctx context.Context, envName string, data runtime.RuntimeEnvironmentData) []string {
```
zombiezen, line 59: Seems like this function should now return an error because it can fail.

workspace/build_manifest.go
```diff
@@ -59,41 +58,53 @@ type ExecPhase struct {
...
 			if _, _, ok := plumbing.SaneEnvironmentVar(property); ok {
-				result = append(result, property)
+				if key, value, ok := plumbing.SaneEnvironmentVar(property); ok {
+					envMap[key] = value
+				}
 			}
 		}
 	}
 
 	// Check for local .env file
```
zombiezen, line 91: Can you help me understand what this previous code didn't do?

workspace/build_manifest.go
```diff
@@ -59,41 +58,53 @@ type ExecPhase struct {
...
 
 	for _, property := range e.Environment["default"] {
 
 		if _, _, ok := plumbing.SaneEnvironmentVar(property); ok {
 			interpolated, err := TemplateToString(property, data)
 			if err == nil {
-				result = append(result, interpolated)
-			} else {
-				result = append(result, property)
+				if key, value, ok := plumbing.SaneEnvironmentVar(interpolated); ok {
```
zombiezen, line 68: The double usage of `SaneEnvironmentVar` is confusing here. Part of the difficulty is its name: it's not evocative of the fact that it splits the string. I recommend renaming the function and either add comments to describe the purpose of this loop, or rework to have less conditionals.

workspace/build_target_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"testing"
+
+	"github.com/yourbase/yb/runtime"
+)
+
+func TestBuildTarget_EnvironmentVariables(t *testing.T) {
+	type fields struct {
+		Name        string
+		Tools       []string
+		Commands    []string
```
zombiezen, line 18: Can we remove these fields? Are they significant for the test?

workspace/build_manifest_test.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/yourbase/yb/runtime"
+)
+
+func TestExecPhase_EnvironmentVariablesDotEnv(t *testing.T) {
+	type fields struct {
+		Name         string
+		Dependencies ExecDependencies
+		Container    narwhal.ContainerDefinition
+		Commands     []string
+		Ports        []string
```
zombiezen, line 22: Are these fields necessary for the test? Can we remove them?

workspace/build_manifest_test.go
```diff
@@ -0,0 +1,134 @@
...
+		Environment  map[string][]string
+		HostOnly     bool
+	}
+	type args struct {
+		ctx     context.Context
+		envName string
+		data    runtime.RuntimeEnvironmentData
+	}
+	// Sets up dotenv
```
zombiezen, line 31: See my comments in the other test file about trying to remove this code.

workspace/build_manifest_test.go
```diff
@@ -0,0 +1,134 @@
...
+		Name         string
+		Dependencies ExecDependencies
+		Container    narwhal.ContainerDefinition
+		Commands     []string
+		Ports        []string
+		Environment  map[string][]string
+		HostOnly     bool
+	}
+	type args struct {
+		ctx     context.Context
```
zombiezen, line 27: Rather than placing the Context in the table test data, pass it in as part of the test loop.

workspace/build_target.go
```diff
@@ -56,17 +57,36 @@ func (b BuildDependencies) ContainerList() []narwhal.ContainerDefinition {
...
+	envMap := make(map[string]string)
+
 	for _, property := range bt.Environment {
-		if _, _, ok := plumbing.SaneEnvironmentVar(property); ok {
-			interpolated, err := TemplateToString(property, data)
-			if err == nil {
-				result = append(result, interpolated)
-			} else {
-				result = append(result, property)
+		interpolated, err := TemplateToString(property, data)
```
zombiezen, line 63: This seems to repeat the code from build_manifest.go quite a bit. Is there a way we can factor out a common function?

zombiezen (commented):
Thanks for working on this feature! I'm excited to have this available, and I appreciate the unit tests.

I'm with you on the direction of this, but I'm having difficulty following the code in this PR. I've left feedback where I saw places to improve the clarity of the code or remove boilerplate. I know there's a lot of comments, but I'm trying to reduce the number of round trips.

workspace/build_manifest.go
beholders-eye: Oh, it didn't override an already set environment variable, so we ended up with duplicated vars, each with different values. This is breaking some of our own projects, in some build targets (e.g: the API). I'll improve comments, and probably improve the logic itself, too many loops.

workspace/build_manifest.go
beholders-eye: Will do

workspace/build_manifest.go
beholders-eye: Makes some sense

workspace/build_manifest_test.go
beholders-eye: OK!

workspace/build_manifest_test.go
beholders-eye: OK!

workspace/build_target.go
beholders-eye: OK!

workspace/build_target_test.go
beholders-eye: Man, that's nice!!

workspace/build_target_test.go
beholders-eye: Good call, yeah!

workspace/build_target_test.go
beholders-eye: OK!

workspace/environment.go
```diff
@@ -0,0 +1,83 @@
...
+func checkAndSplitEnvVar(env string) (name, value string, sane bool) {
+	s := strings.SplitN(env, "=", 2)
+	if sane = (len(s) == 2); sane {
+		name = s[0]
+		value = s[1]
+	}
+	return
+}
+
+// parseEnvironment sanely checks and process an arbitrary number of
```
zombiezen, line 28: nit: "sanely" doesn't add anything descriptive to the documentation and furthers stigmatization of mental health issues.

```suggestion
// parseEnvironment checks and process an arbitrary number of
```

zombiezen (approved):
LGTM after one nit:

workspace/environment.go
beholders-eye: Oh, sorry. Thanks for pointing this out.

PR #169, opened 2020-09-10
UNIX symbols from 20200831 wasn't working

beholders-eye:
When build for GOOS=darwin


beholders-eye:
Found this by running `local_test_release` locally.

PR #175, opened 2020-09-21
Remove dead code in plumbing and unused server package

zombiezen:


beholders-eye (approved):
Thanks for clearing this up. The server was still around because we were planing a new feature: telemetry. But thanks!!

PR #176, opened 2020-09-21
Remove dot imports

zombiezen:
https://github.com/golang/go/wiki/CodeReviewComments#import-dot

### DIFF
--- a/buildpacks/anaconda.go
+++ b/buildpacks/anaconda.go
@@ -6,14 +6,13 @@ import (
 	"path/filepath"
 
 	"github.com/blang/semver"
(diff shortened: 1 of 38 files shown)


beholders-eye (approved):
This is straightforward but touches a lot of sources. I really like it, make everything more readable. I'd like to know if you used some tool to do it. Did you?

zombiezen:
Nothing terribly automatic, but my process was to remove the dots from the top of the file and then in VSCode repeatedly use F8 to jump to next compile error and Ctrl-Space to autocomplete the appropriate function call.

PR #177, opened 2020-09-21
Fix problems found by staticcheck

zombiezen:


buildpacks/python.go
```diff
@@ -13,7 +13,7 @@ import (
 const (
 	anacondaToolVersion = "4.8.3"
 	// The version above needs a newer template
-	anacondaURLTemplate = "https://repo.continuum.io/miniconda/Miniconda{{.PyNum}}-py37_{{.Version}}-{{.OS}}-{{.Arch}}.{{.Extension}}"
+	anacondaURLTemplate = "https://repo.continuum.io/miniconda/Miniconda{{.PyNum}}-{{.Version}}-{{.OS}}-{{.Arch}}.{{.Extension}}"
```
beholders-eye, line 16: This will break our Python build pack. It needs that "-py37_", why did it remove it?

plumbing/util.go
```diff
@@ -115,39 +114,6 @@ func ExecSilentlyToWriter(cmdString string, targetDir string, writer io.Writer)
 
 }
 
-func ExecToLogWithProgressDots(cmdString string, targetDir string, logPath string) error {
```
beholders-eye, line 118: This was mostly a "mark" to remind us that we should use some better terminal handling library/tool to show some progress. I've added something in the remote build code path for it, but it's not optimal.

cli/run.go
```diff
@@ -13,7 +13,6 @@ import (
 )
 
 type RunCmd struct {
-	target      string
```
beholders-eye, line 16: We do use a `target` in runtime-context...

cli/remote_build.go
```diff
@@ -159,9 +159,7 @@ func (p *RemoteCmd) Execute(_ context.Context, f *flag.FlagSet, _ ...interface{}
 
 	for _, r := range list {
 		c := r.Config()
-		for _, u := range c.URLs {
-			repoUrls = append(repoUrls, u)
-		}
+		repoUrls = append(repoUrls, c.URLs...)
```
beholders-eye, line 162: This one is useful...

cli/login.go
```diff
@@ -70,6 +70,10 @@ func (p *LoginCmd) Execute(_ context.Context, f *flag.FlagSet, _ ...interface{})
 	}
 
 	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
```
beholders-eye, line 73: This fix is in runtime-context, IIRC

buildpacks/properties_linux.go
```diff
@@ -42,5 +42,4 @@ func getDebianOrUbuntuVersion() string {
 			}
 		}
 	}
-	return ""
```
beholders-eye, line 45: This code is being completely removed in #173

beholders-eye (changes requested):
The fix for `buildpacks/python.go` will break how we download Anaconda/Miniconda.

buildpacks/python.go
zombiezen: staticcheck (correctly) reports that this constant isn't being used, so I'm replacing it with what is actually being used.

cli/login.go
zombiezen: Acknowledged.

cli/run.go
zombiezen: Acknowledged, but it's not currently in-use in this branch.

plumbing/util.go
zombiezen: Understood, but it's unused code and not properly synchronized. Rather than fixing it, I'm opting to remove the code.

Context: staticcheck is warning that a `for/select/default` combo can cause busy-waits. This is true in general, but the `default` case includes a `time.Sleep`. However, the code doesn't use a `time.Ticker` or other such to permit `stoppedchan` to interrupt the sleep, so the goroutine may leak past the end of the overall function.

buildpacks/python.go
johnewart: Given that the old uppercase URL wasn't correct either I'm not 100% sure how this works now -- though maybe it doesn't (we should test this). There's a -latest, it might be easiest to just use that for now in the `anacondaToolVersion` (it seems their previous naming scheme was just the version up until 4.8)

johnewart (commented):
LGTM

PR #180, opened 2020-09-24
Clean up plumbing.IsBinary

zombiezen:
Roll-forward of d176099f6dbf942040f113672cf89d25e8453bba (#119)


plumbing/util.go
```diff
@@ -459,36 +457,27 @@ func CloneRepository(remote types.GitRemote, inMem bool, basePath string) (rep *
...
+	for _, b := range buf[:n] {
+		if b == 0 {
+			return true, err
 		}
-
-		c++
 	}
-
-	return false, nil
+	return false, err
```
glacials, line 482: This function is just a little hard for me to follow because of how `err` can waterfall down in different ways, but considering how short it is it's not blocking at all and could just be a difference in taste.

glacials (approved):

plumbing/util.go
zombiezen: Thanks for pointing that out! I added a comment because the `io.Reader` "return both data and an error sometimes" behavior is definitely confusing and contrary to most other Go APIs.

PR #181, opened 2020-09-24
Use OpenTelemetry for timing information

zombiezen:
Fixes ch-2530


cli/build.go
```diff
@@ -15,6 +15,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/johnewart/archiver"
```
johnewart, line 21: For reference, this is only because I have a patch that fixes tar files with hard links, which I think Flutter had an issue with

johnewart (approved):
:tada:

cli/build.go
zombiezen: Opened https://app.clubhouse.io/yourbaseio/story/2561/use-upstream-archiver-in-cli

PR #183, opened 2020-09-24
Clean up redundant code in Homebrew buildpack

zombiezen:


buildpacks/homebrew.go
```diff
@@ -106,38 +103,53 @@ func (bt HomebrewBuildTool) InstallDir() string {
...
+	if _, err := os.Stat(brewDir); err == nil {
+		log.Infof("brew installed in %s", brewDir)
+		return nil
 	}
 
+	log.Infof("Installing brew")
+	installDir := bt.InstallDir()
+	if err := os.MkdirAll(installDir, 0777); err != nil {
+		return fmt.Errorf("install homebrew: %w", err)
+	}
```
glacials, line 136: Do we really need `0777`?

buildpacks/homebrew.go
```diff
@@ -106,38 +103,53 @@ func (bt HomebrewBuildTool) InstallDir() string {
...
-		} else {
-			log.Infof("Installing package %s", bt.PackageVersionString())
-			err = bt.InstallPackage()
-			if err != nil {
-				return err
-			}
-		}
+	log.Infof("Updating brew")
+	if err := plumbing.ExecToStdout("brew update", brewDir); err != nil {
+		return fmt.Errorf("install homebrew: %w", err)
```
glacials, line 147:
```suggestion
		return fmt.Errorf("brew update: %w", err)
```

glacials (approved):

buildpacks/homebrew.go
zombiezen: Why not both? 😉 Done.

buildpacks/homebrew.go
zombiezen: So generally, unless there's a specific reason to restrict permissions, it is better to use `0666` or `0777` as modes to respect the user's [umask](https://en.wikipedia.org/wiki/Umask). `0022` is a fairly common umask value, which is why most programs automatically do "the right thing".

glacials: Ah, didn't know about umask. Thanks for the link!

PR #184, opened 2020-09-24
Replace MkdirAsNeeded with os.MkdirAll

zombiezen:
[`os.MkdirAll`](http://pkg.go.dev/os#MkdirAll) behaves identically.

### DIFF
--- a/buildpacks/glide.go
+++ b/buildpacks/glide.go
@@ -81,7 +81,9 @@ func (bt GlideBuildTool) Install() error {
 		}
 
 		extractDir := bt.GlideDir()
-		plumbing.MkdirAsNeeded(extractDir)
+		if err := os.MkdirAll(extractDir, 0777); err != nil {
+			return err
+		}
 		log.Infof("Extracting glide %s to %s...", bt.Version(), extractDir)
 		err = archiver.Unarchive(localFile, extractDir)
 		if err != nil {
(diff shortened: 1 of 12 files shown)


buildpacks/glide.go
```diff
@@ -81,7 +81,9 @@ func (bt GlideBuildTool) Install() error {
 		}
 
 		extractDir := bt.GlideDir()
-		plumbing.MkdirAsNeeded(extractDir)
+		if err := os.MkdirAll(extractDir, 0777); err != nil {
+			return err
```
glacials, line 85: Consistency with those below:
```suggestion
			return fmt.Errorf("install Glide: %w", err)
```

plumbing/util.go
```diff
@@ -230,7 +222,9 @@ func CacheDir() string {
 		}
 	}
 
-	MkdirAsNeeded(cacheDir)
+	if err := os.MkdirAll(cacheDir, 0777); err != nil {
+		log.Errorf("%v", err)
+	}
```
glacials, line 227: For each of the three `os.MkdirAll`s in this file, how fantastically do we expect the rest of the invocation to blow up if we reach errors? Is it worth plumbing them through to their callers?

johnewart (approved):

glacials (approved):

plumbing/util.go
zombiezen: Opened ch-2567

PR #185, opened 2020-09-24
Remove progress spinners

zombiezen:
The code to render the spinners was improperly copied into yb without the correct licensing information: it should be added as a dependency or properly sectioned off if we were to add it back.

It was only being used in remote build, and I'd like to revisit the output in a more holistic take.


cli/remote_build.go
```diff
@@ -272,22 +235,15 @@ func (p *RemoteCmd) Execute(_ context.Context, f *flag.FlagSet, _ ...interface{}
...
 			return subcommands.ExitFailure
 		}
 
-		if log.CheckIfTerminal() {
-			patchProgress = plumbing.NewProgressSpinner("Generating patch for local changes")
-			patchProgress.Start()
-		} else {
-			log.Info("Generating patch for local changes")
-		}
+		log.Info("Generating patch for local changes")
```
glacials, line 242: Just for consistency with the above output:
```suggestion
		log.Info("Generating patch for local changes...")
```

glacials (approved):

PR #187, opened 2020-09-24
Unexport most of buildpacks package

zombiezen:
Other parts of the codebase only use buildpacks to install a buildpack, so I've unexported these parts to make their purpose clearer and unlock more aggressive refactorings in the future.


jamesnaftel (approved):
LGTM.

PR #188, opened 2020-09-25
Add `yb token` command

glacials:
When developing and testing the API you generally need an auth token. The auth tokens are long enough to cause a single `curl` to span many lines, making it unwieldy to deal with requests. By adding a `yb token` command and inserting it into curl we can get something like
```sh
curl -H "Authorization: Bearer `yb token`" https://api.yourbase.io/users/me
```
instead of something like
```sh
curl -H 'Authorization: Bearer 123bh123bh79h12b1237123hv12b123bvn12873bvn10293vn8127v0129bv3n120v91cv12nbccbn2106bncbvc123bcv201bbcnb12987bncx9128730cb1n2xbnv2523v2v2v23v523v23v23c.v.ewfhwevj1234n9-cv192bnc12blwafvwvfbnvwfab09nwva0bnv9ew8abn8vf09aw6bvwa' https://api.yourbase.io/users/me
```

### DIFF
--- a/cli/token.go
+++ b/cli/token.go
@@ -1,0 +1,49 @@
+package cli
+
+import (
+	"context"
+	"flag"
+	"fmt"
+
+	"github.com/johnewart/subcommands"
+	"github.com/yourbase/yb/config"
+	"github.com/yourbase/yb/plumbing/log"
+)
+
+// TokenCmd represents an invocation of `yb token`, which outputs the saved
+// token from `yb login` to stdout. This can be used
+type TokenCmd struct {
+	Version string
+	Channel string
+}
+
+// Name returns the literal text of the token command.
+func (*TokenCmd) Name() string { return "token" }
+
+// Synopsis returns the shortform description of the token command.
+func (*TokenCmd) Synopsis() string {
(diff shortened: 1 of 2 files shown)


cli/token.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/johnewart/subcommands"
+	"github.com/yourbase/yb/config"
+	"github.com/yourbase/yb/plumbing/log"
+)
+
+// TokenCmd represents an invocation of `yb token`, which outputs the saved
+// token from `yb login` to stdout. This can be used
+type TokenCmd struct {
+	Version string
+	Channel string
```
zombiezen, line 17: IIUC these fields don't get used. Delete?

zombiezen (approved):

zombiezen:
IIUC this addresses ch-2516

cli/token.go
glacials: Oops, copy-paste cruft. Nice catch!

glacials:
Correct! Let's see if it auto-updates if not mentioned by me specifically.

PR #190, opened 2020-09-25
Switch to zombiezen.com/go/log

zombiezen:
Along the way, cleaned up the `managementLogUrl` function and removed some unused patch-related functions.

Fixes ch-2528


cli/config.go
```diff
@@ -89,7 +89,7 @@ func (w *configGetCmd) SetFlags(f *flag.FlagSet) {}
 
 func (w *configGetCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface{}) subcommands.ExitStatus {
 	if len(f.Args()) < 1 {
-		log.Warnln(w.Usage())
+		log.Warnf(ctx, "%s", w.Usage())
```
glacials, line 92: I have no preference either way, but any reason for not doing a
```suggestion
		log.Warnf(ctx, w.Usage())
```
?

cli/remote_build.go
```diff
@@ -187,87 +186,84 @@ func (p *RemoteCmd) Execute(_ context.Context, f *flag.FlagSet, _ ...interface{}
...
+			log.Errorf(ctx, "%s", err)
 		}
 
 		// TODO When go-git worktree.Status() get faster use this instead:
 		// There's also the problem detecting CRLF in Windows text/source files
 		//if err = worktree.AddGlob("."); err != nil {
-		if skippedBinaries, err := p.traverseChanges(worktree, saver); err != nil {
-			log.Error(err)
+		if skippedBinaries, err := p.traverseChanges(ctx, worktree, saver); err != nil {
+			log.Errorf(ctx, "%v", err)
```
glacials, line 246: On line 239 we use `%s`, here we use `%v`, do we want to make these consistent?

cli/remote_build.go
```diff
@@ -570,44 +566,40 @@ func postToApi(path string, formData url.Values) (*http.Response, error) {
...
+	// Pattern is /builds/ID/progress
+	const prefix = "/builds/"
+	const suffix = "/progress"
+	if !strings.HasPrefix(u.Path, prefix) || !strings.HasSuffix(u.Path, suffix) {
+		return "", fmt.Errorf("build ID for %v: unrecognized path", u)
+	}
+	id := u.Path[len(prefix) : len(u.Path)-len(suffix)]
+	for i := 0; i < len(id); i++ {
+		if id[i] == '/' {
+			return "", fmt.Errorf("build ID for %v: unrecognized path", u)
```
glacials, line 582: Any reason not to use `strings.ContainsRune` here?

cli/remote_build.go
```diff
@@ -729,76 +721,79 @@ func (cmd *RemoteCmd) submitBuild(project *types.Project, tagMap map[string]stri
...
+	if err != nil {
+		return fmt.Errorf("server response: parse log URL: %w", err)
+	}
+	if logURL.Scheme != "ws" && logURL.Scheme != "wss" {
+		return fmt.Errorf("server response: parse log URL: unhandled scheme %q", logURL.Scheme)
+	}
+	uiURL := ""
+	if id, err := buildIDFromLogURL(logURL); err == nil {
+		// Fine to ignore error: this is displayed as a fallback if other things fail.
+		uiURL, _ = ybconfig.ManagementUrl("/" + project.OrgSlug + "/" + project.Label + "/builds/" + id)
```
glacials, line 739: I agree it's fine to continue if `err != nil`, but we could get some value out of outputting it anyway, so we don't overlook something silently failing.

glacials (approved):
2dd

cli/config.go
zombiezen: If the usage string contains a percentage sign, then it would be interpreted as a format code. Using `"%s"` ensures the string is interpreted literally.

cli/remote_build.go
zombiezen: Done.

cli/remote_build.go
zombiezen: Good call. Done.

cli/remote_build.go
zombiezen: Done.

PR #191, opened 2020-09-29
Add GitHub Actions-based release process

zombiezen:
I'm using GitHub Actions primarily for macOS support: I firmly believe we should be running our tests on macOS, not just cross-compiling at release. This change retains the YourBase CI for main builds, but removes the release targets.

I restructured the release automation scripts to be more modular and have more well-defined outputs. I've removed the Equinox process, since I'd like to revisit our packaging more broadly in future PRs.

Fixes ch-2032


jamesnaftel (approved):
LGTM.   I like the changes.   The commands list was looking a little unwieldy in .yourbase.yml.

PR #192, opened 2020-09-29
Add Debian binary packaging step to release process

zombiezen:
Bypasses the Debian source packaging route, since we have a lot of Go dependencies and we don't want to package them all for Debian.

Fixes ch-2577


jamesnaftel (approved):
LGTM.

PR #196, opened 2020-10-05
Upgrade yb to use latest Narwhal

zombiezen:
Fixes ch-1470


jamesnaftel (approved):
LGTM.  I like the new friend error messages link the docs!

zombiezen:
macOS build failure is legitimate, sadly. Seems that the Docker client pulls in `github.com/docker/docker/pkg/term` which depends on an unsupported Darwin syscall mechanism for ioctl. Investigating how to work around...

PR #197, opened 2020-10-06
Fix Python buildpack Anaconda download URL

zombiezen:
Use same logic as in Anaconda buildpack and add unit tests to verify. Tweaked Python buildpack to properly disable interactive prompt.

Fixes ch-2671


glacials (approved):
Nice tests!

PR #198, opened 2020-10-06
Bail early on HTTP downloads for non-200 responses

zombiezen:
Also add tracing and cancellation. Now using a single tracer for entire project.

Fixes ch-2672


zombiezen:
I have two somewhat unrelated commits in this PR, but this PR depends on the first. LMK if you want me to split PRs.

glacials (approved):
👌 Nah you're good.

PR #200, opened 2020-10-08
Respect XDG variables for configuration files

zombiezen:
Reworked the configuration package to match best practices and reduce the number of filesystem accesses.

Updates ch-1401


config/config_file.go
```diff
@@ -1,109 +1,84 @@
...
+	settingsName = "settings.ini"
+)
 
-	if !pathExists(iniPath) {
-		emptyFile, _ := os.Create(iniPath)
-		emptyFile.Close()
+func loadConfigFiles() (*ini.File, error) {
+	// Gather INI files in reverse order because earlier entries should override
+	// later entries.
+	searchPaths := xdgdir.Config.SearchPaths()
```
glacials, line 23: After reading this comment I believed "earlier"/"later" referred to timestamp, which was confusing enough that I was compelled to dig into `xdgdir` and `ini` docs to unravel the mystery. Can we clarify this comment a bit to explain it upfront?
```suggestion
	// xdgdir.Config.SearchPaths returns config files in descending order of preference,
	// but ini.Load accepts them in ascending order of preference, so gather them in reverse.
	searchPaths := xdgdir.Config.SearchPaths()
```
(feel free to edit)

config/config_file.go
```diff
@@ -1,109 +1,84 @@
...
-	return nil
+	return name
 }
 
-func SectionPrefix() string {
-	profile := YourBaseProfile()
-	if profile != "" {
-		return fmt.Sprintf("%s:", profile)
+func Get(section string, key string) (string, error) {
+	cfg, err := loadConfigFiles()
```
glacials, line 55: Are we reading from disk every time `Get` is called? If so, can we cache the read in memory?

This seems unchanged from current behavior so I'd be fine if it's in a separate PR.

config/settings.go
```diff
@@ -4,59 +4,41 @@ import (
...
+func apiBaseURL() (string, error) {
 	if url, exists := os.LookupEnv("YOURBASE_API_URL"); exists {
 		return url, nil
 	}
-
-	if url, err := GetConfigValue("yourbase", "api_url"); err == nil {
-		if url != "" {
-			return url, nil
-		}
+	cfg, err := loadConfigFiles()
```
glacials, line 34: Same question about reading from disk.

config/settings.go
```diff
@@ -68,42 +50,34 @@ func apiBaseUrl() (string, error) {
...
+func uiBaseURL() (string, error) {
+	if url := os.Getenv("YOURBASE_UI_URL"); url != "" {
 		return url, nil
 	}
-
-	if url, err := GetConfigValue("yourbase", "management_url"); err == nil {
-		if url != "" {
-			return url, nil
-		}
+	cfg, err := loadConfigFiles()
```
glacials, line 73: Same question about reading from disk.

config/settings.go
```diff
@@ -115,58 +89,50 @@ func managementBaseUrl() (string, error) {
...
 }
 
-func CurrentGHAppUrl() (gh string) {
-	profile := YourBaseProfile()
-
+func GitHubAppURL() (gh string) {
+	profile := "production"
+	if cfg, err := loadConfigFiles(); err == nil {
+		profile = getProfile(cfg)
+	}
```
glacials, line 112: Same question about reading from disk.

config/settings.go
```diff
@@ -68,42 +50,34 @@ func apiBaseUrl() (string, error) {
...
-func ApiUrl(path string) (string, error) {
-	if !strings.HasPrefix(path, "/") {
-		path = fmt.Sprintf("/%s", path)
-	}
-
-	if baseUrl, err := apiBaseUrl(); err != nil {
-		return "", fmt.Errorf("Can't determine API URL: %v", err)
-	} else {
-		return fmt.Sprintf("%s%s", baseUrl, path), nil
+func APIURL(path string) (string, error) {
```
glacials, line 57: This is unchanged in this PR, but a nice cleanup here could be to make all these `*URL` methods return `url.URL`s instead, for safety and sanity.

glacials (approved):
Nothing blocking, mostly comments on supplementary cleanup. Nice refactors, and today I learned about XDG!

config/settings.go
zombiezen: Created ch-2692 to track.

config/config_file.go
zombiezen: Sadly, yes, it is making syscalls every time. I would like to cache the read in-memory, but I was worried in general about the interaction with `Set()`, even though the calls I've seen so far seem okay. I'm relying on the fact that the files are small enough that the OS will likely keep them in cache rather than actually hit disk to handle the overall synchronization.

Opened ch-2693 to track cleanup.

PR #201, opened 2020-10-08
Remove ToolsDir

zombiezen:
The buildpacks are now per-target and will likely need to be more fine-grained over time. Storing the data along with the target that uses it increases reproducibility and simplifies cleanup at the cost of download time and disk space. The former can be addressed with judicious use of `DownloadFileWithCache`.

Updates ch-1401
Updates ch-2567

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ The format is based on [Keep a Changelog][], and this project adheres to
 
 ## [Unreleased][]
 
+### Changed
+
+-  Buildpacks are no longer installed to `$HOME/.yourbase/tools`. They are now
+   installed into the per-build context cache. You can safely remove
+   `$HOME/.yourbase/tools` to reclaim disk space.
+
 ### Fixed
 
 -  Update to latest version of [Narwhal](https://github.com/yourbase/narwhal),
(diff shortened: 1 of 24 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ The format is based on [Keep a Changelog][], and this project adheres to
 
 ## [Unreleased][]
 
+### Changed
+
+-  Buildpacks are no longer installed to `$HOME/.yourbase/tools`. They are now
+   installed into the per-build context cache. You can safely remove
+   `$HOME/.yourbase/tools` to reclaim disk space.
+
```
glacials, line 16: Any reason we can't remove this for them?

glacials (approved):

CHANGELOG.md
zombiezen: I'm always a little nervous about tools that delete without user interaction (what if you still have an old version of yb hanging around? etc.). We could check for the directory on startup and log a message?

zombiezen: (I'm going to merge this to unblock some follow-on PRs I had queued, but I'm happy to come back and address this if you want it included.)

glacials: If you think it's generally going to be < 1 GB then it's probably not worth the prompt. If not, I think it is worth it to be a good neighbor rather than silently eating space for potentially years.

zombiezen: Opened ch-2788

PR #202, opened 2020-10-08
Remove extract-to-disk step of yb download for containers

zombiezen:
Eliminates a dependency on the `CacheDir`. `DownloadYB` will be eventually removed anyway, but we need to keep it around for a little while.

Updates ch-1401

### DIFF
--- a/cli/build.go
+++ b/cli/build.go
@@ -771,7 +771,7 @@ func sha256File(path string) (string, error) {
 
 // TODO: non-linux things too if we ever support non-Linux containers
 // TODO: on non-Linux platforms we shouldn't constantly try to re-download it
-func DownloadYB(ctx context.Context) (_ io.ReadCloser, size int64, _ error) {
+func DownloadYB(ctx context.Context) (_ io.ReadCloser, size int64, err error) {
 	// Stick with this version, we can track some relatively recent version because
 	// we will just update anyway so it doesn't need to be super-new unless we broke something
 	const downloadURL = "https://bin.equinox.io/a/7G9uDXWDjh8/yb-0.0.39-linux-amd64.tar.gz"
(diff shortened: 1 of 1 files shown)


cli/build.go
```diff
@@ -434,40 +436,51 @@ func (b *BuildCmd) BuildInsideContainer(ctx context.Context, target *types.Build
...
 	// Upload and update CLI
-	log.Infof(ctx, "Uploading YB from %s to /yb", localYbPath)
-	err = narwhal.UploadFile(ctx, dockerClient, buildContainer.Id, "/yb", localYbPath)
+	log.Infof(ctx, "Uploading YB to /yb")
+	err = narwhal.Upload(ctx, dockerClient, buildContainer.Id, "/yb", localYB, &tar.Header{
+		Mode:     0777,
+		Size:     localYBSize,
+		Typeflag: tar.TypeReg,
+	})
+	localYB.Close() // just a reader, Close can be ignored.
```
glacials, line 483: Would this be more idiomatic in a `defer` near declaration or assignment?

cli/build.go
```diff
@@ -758,40 +771,65 @@ func sha256File(path string) (string, error) {
...
 	}
+	if !found {
+		zr.Close()
+		archiveFile.Close()
+		return nil, 0, fmt.Errorf("download yb: no 'yb' found in %s", downloadURL)
+	}
+	return multiCloseReader{
+		Reader:  archive,
+		closers: []io.Closer{zr, archiveFile},
+	}, size, nil
```
glacials, line 817: Can this code have the `found` and `size` pieces of state simplified away by moving this `return` up into the loop?
```suggestion
	for {
		hdr, err := archive.Next()
		if errors.Is(err, io.EOF) {
			break
		}
		if err != nil {
			zr.Close()
			archiveFile.Close()
			return nil, 0, fmt.Errorf("download yb: %s: %w", archivePath, err)
		}
		if hdr.Name == "yb" {
			return multiCloseReader{
				Reader:  archive,
				closers: []io.Closer{zr, archiveFile},
			}, hdr.Size, nil
		}
	}
	zr.Close()
	archiveFile.Close()
	return nil, 0, fmt.Errorf("download yb: no 'yb' found in %s", downloadURL)

```

cli/build.go
```diff
@@ -758,40 +771,65 @@ func sha256File(path string) (string, error) {
...
+		return nil, 0, fmt.Errorf("download yb: %w", err)
 	}
-
-	// Couldn't tell, check if we need to and download the archive
-	localFile, err := plumbing.DownloadFileWithCache(ctx, http.DefaultClient, downloadURL)
+	archiveFile, err := os.Open(archivePath)
 	if err != nil {
-		return "", fmt.Errorf("download yb: %w", err)
+		return nil, 0, fmt.Errorf("download yb: %w", err)
 	}
```
glacials, line 785: Can we simplify away the scattered `archiveFile.Close()`s below with something like
```go
defer func() {
  if err != nil {
    archiveFile.Close()
  }
}()
```
?

cli/build.go
```diff
@@ -758,40 +771,65 @@ func sha256File(path string) (string, error) {
...
-
-	err = archiver.Unarchive(localFile, tmpPath)
+	zr, err := gzip.NewReader(archiveFile)
 	if err != nil {
-		return "", fmt.Errorf("download yb: decompress: %v", err)
+		archiveFile.Close()
+		return nil, 0, fmt.Errorf("download yb: %s: %w", archivePath, err)
+	}
```
glacials, line 790: Same `defer` question for `zr`

glacials (approved):

cli/build.go
zombiezen: I was hesitant on this one because the containing function is large, but I solved by extracting this to a function.

PR #203, opened 2020-10-09
Pass in data directories as an explicit input

zombiezen:
Makes it easier to override in tests and separates missing environment variable errors from path generation.

Fixes ch-2567
Updates ch-1401


zombiezen:
Sorry, wanted to add a little more scope here. Will come back with a new PR.

zombiezen:
Obsoleted by #204

PR #204, opened 2020-10-09
Pass in data directories as an explicit input

zombiezen:
Makes it easier to override in tests and separates missing environment variable errors from path generation.

Moved build roots to be created underneath `$XDG_CACHE_HOME`.

Fixes ch-1401
Fixes ch-2567

### DIFF
--- a/internal/ybdata/ybdata.go
+++ b/internal/ybdata/ybdata.go
@@ -58,7 +58,7 @@ func DirsFromEnv() (*Dirs, error) {
 func NewDirs(root string) *Dirs {
 	return &Dirs{
 		cache:      root,
-		workspaces: filepath.Join(cache, "workspaces"),
+		workspaces: filepath.Join(root, "workspaces"),
 	}
 }
 
--- a/internal/ybdata/ybdata.go
+++ b/internal/ybdata/ybdata.go
@@ -32,8 +32,8 @@ type Dirs struct {
 	workspaces string
 }
 
-// FromEnv finds data directories based on environment variables.
-func FromEnv() (*Dirs, error) {
+// DirsFromEnv finds data directories based on environment variables.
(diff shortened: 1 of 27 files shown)


internal/ybdata/ybdata.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+// Dirs is the result of locating directories.
+type Dirs struct {
+	cache      string
+	workspaces string
+}
+
+// FromEnv finds data directories based on environment variables.
+func FromEnv() (*Dirs, error) {
```
glacials, line 36: Based on Effective Go's [_Package Names_][1] section the name `ybdata.FromEnv` makes me believe `FromEnv` should return "a ybdata". I think this could flow better if the function was called `DirsFromEnv`, or if the package was renamed to `dirs` or `ybdirs` so you could call `dirs.New` and `dirs.FromEnv`.

[1]: https://golang.org/doc/effective_go.html#package-names

internal/ybdata/ybdata.go
```diff
@@ -0,0 +1,75 @@
...
+	"path/filepath"
+
+	"go4.org/xdgdir"
+)
+
+// Dirs is the result of locating directories.
+type Dirs struct {
+	cache      string
+	workspaces string
+}
```
glacials, line 33: Any particular reason we are passing around `*Dirs` everywhere instead of `Dirs`? It's a fairly light struct and I don't see any case where it's intended to be nil, so would it be worth it to gain the safety of not dealing with pointers?

glacials (approved):
I like this refactor, but what's your thinking for where the responsibility of creating these directories should lie long-term? I see some consumers fetching the directory and then immediately `os.MkdirAll`ing it, and some assuming that it exists.

Are we leaving it up to the consumer to decide whether it's safe to use a directory or not?

PR #211, opened 2020-10-12
Add concepts reference

zombiezen:
This documents the near-future target for this tool, not the tool as it exists today.

Fixes ch-2696

### DIFF
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -1,56 +1,93 @@
-# yb Concepts
+# yb Concept Reference
+
+This document describes the semantics and terms used by yb. It is aimed at
+developers and advanced users of yb, and can be seen as a light-weight
+specification of a _build runner_ &mdash; the yb CLI or any other system that
+runs a YourBase build.
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in [RFC 2119][].
+
+[RFC 2119]: https://tools.ietf.org/html/rfc2119
+
+## Targets
+
+A _target_ is a buildable thing &mdash; it could be a service, a library,
+a binary, etc. Targets can depend on zero or more other targets, creating a
+[directed acyclic graph][].
+
+A target has five _phases_, each of which has a list of one or more _commands_
+to run. A command is a program invocation, like `gcc -o foo foo.c` or
+`npm install`. These phases are all optional, but at least one **MUST** be
+specified per target. The phases of a target are:
+
+-  **Build** **MUST** be run when the target is depended on and before any of
+   the other phases in the same target. This phase can optionally declare output
+   files, which **SHALL** be accessible to other targets that depend on this
+   target and any other phases in the same target.
+-  **Test** _SHOULD_ run the target's test suite. Common commands in this stage
+   would include `go test ./...` or `npm test`.
+-  **Execute** **MUST** be run when the target is used as a resource for
+   another target (discussed later in this section).
+-  **Release** _SHOULD_ create release output files, like tarballs or binary
+   packages. This phase can optionally declare output files, which **SHALL** be
+   accessible to the deploy phase of the same target.
+-  **Deploy** _SHOULD_ deploy the release output.
+
+The build, test, and release phases _SHOULD_ be safe to be run multiple
+times. Automated builds (like in CI) _MAY_ retry these phases without prompting
+the user. Commands in these phases _MAY_ also be skipped or reordered if the
+build runner detects that doing so won't alter the results of the phase.
+Commands in the execute and deploy phases **MUST NOT** be skipped or retried.
+
+A _resource_ is a process that runs concurrently with a phase's commands.
+This can be either another target's execute phase or it may be a Docker
+container.
+
+**TODO:** Include example of target YAML.
+
+[directed acyclic graph]: https://en.wikipedia.org/wiki/Directed_acyclic_graph
+
+### Commands
+
+Each phase **SHALL** run in a separate, isolated environment. For example,
+a phase may run in a Docker container or a chroot jail. A phase _MAY_ run on
+a different host from the build runner. Commands executed in the phase **MUST**
+have the following environment variables set:
+
+-  `HOME` **MUST** be set to the path of an readable and writable directory.
+   This _SHOULD NOT_ be the same as the user's actual `HOME` directory to
+   keep builds reproducible.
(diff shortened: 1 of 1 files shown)


jamesnaftel:
I like it.

docs/concepts.md
```diff
@@ -0,0 +1,56 @@
...
+## Target
+
+A _target_ is a buildable thing &mdash; it could be a service, a library,
+a binary, etc. A target has four _phases_ (all optional, but at least one must
+be present):
+
+-  Build
+-  Test
+-  Execute
+-  Release
```
glacials, line 20: I'm super on board with being opinionated about these four phases existing as first-class citizens, and not just shoving them all into the "build" cubby 👍

However with _each target_ having each of these four phases, it's unclear to me how we expect customers to draw the lines between them. For example, take a project which in today's world has these targets:
```
seeddb
test
integrationtest
release
```
Plus this "non-target":
```
exec
```

With each target gaining build/test/execute/release, how do we expect this project to split their targets up? The closest I can see is something like
```
seeddb
  build
api
  build
  test
  exec
  release
integration
  test
```

And this seems like it's actually more shoehorned than the current system, not less.

Should we consider build/test/exec/release to be _categories of targets_, such that typing `yb test` runs all targets categorized as tests, `yb exec` runs all targets (e.g. PostgreSQL, Redis, app, etc.) categorized as `exec`, etc.?

As a meta-comment, it would be a good practice to go through some of our customers' build configs and refactor them to how they would look in this new system to reveal any oddness. Happy to pair on some of this for the benefit of creative juices.

docs/concepts.md
```diff
@@ -0,0 +1,56 @@
+# yb Concepts
+
+## Package
+
+A _package_ is a collection of related files and targets. A package is a
+directory that contains a .yourbase.yml file directly inside it. Files can only
+belong to a single package. A package may contain subpackages in subdirectories,
+but files within those subdirectories belong to the subpackage, not the parent
+package.
```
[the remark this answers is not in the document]
glacials: Does a file in a non-package subdirectory belong to the package, or is it packageless? e.g.
```
./app
  .yourbase.yml
  a.txt
  b/
    c.txt
```
Does `c.txt` belong to `app`?

If so, are we okay with that relationship being implicitly destroyed if `b/.yourbase.yml` is created?

docs/concepts.md
```diff
@@ -0,0 +1,56 @@
...
+different host.
+
+Programs in the context are guaranteed to have access to all the files in the
+package and the following environment variables:
+
+-  `HOME`
+-  `LOGNAME`
+-  `PATH`
+-  `TZ=UTC`
+-  `USER`
```
glacials, line 46: Can we get brief descriptions of these?

glacials (changes requested):
👍 Solid first steps

docs/concepts.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Buildpacks
+
+A _buildpack_ installs development tools into a context. It sets up environment
+variables and files before a phase is executed.
+
+## Resources
+
+A _resource_ is a process that runs concurrently with the build. This may be
```
amyhua, line 55: can you elaborate?

also, i may shy away from broadly named names. Something more specific is generally better IMO, because we can run into name pollution with colloqial use, very quickly.

docs/concepts.md
```diff
@@ -0,0 +1,56 @@
...
+A _package_ is a collection of related files and targets. A package is a
+directory that contains a .yourbase.yml file directly inside it. Files can only
+belong to a single package. A package may contain subpackages in subdirectories,
+but files within those subdirectories belong to the subpackage, not the parent
+package.
+
+## Target
+
+A _target_ is a buildable thing &mdash; it could be a service, a library,
+a binary, etc. A target has four _phases_ (all optional, but at least one must
```
amyhua, line 14: (nitpick) --> "A target has 1 or more of the following phases"

docs/concepts.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Buildpacks
+
+A _buildpack_ installs development tools into a context. It sets up environment
+variables and files before a phase is executed.
+
+## Resources
+
+A _resource_ is a process that runs concurrently with the build. This may be
+another target's execute phase or it may be a Docker container.
```
amyhua, line 56: now that i've read the whole thing --- i think, for kind of a sanity check on the doc as a whole, specifying a diagram of simply (without concepts) what is all required, generally, in the whole build/deployment process, and then, mapping those needs to these concepts and how these concepts get applied, would be a nice sanity check and help me with establishing a clearer model of how these concepts all fit as a whole, to the general CI/CD experience.

amyhua (changes requested):
This is a nice initial outline!

I left a number of comments, mostly asking for clarity before moving on to give my thoughts on them.

I propose also specifying the phases more, and reduce their number to 3 for easier picking for the customer.

docs/concepts.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Target
+
+A _target_ is a buildable thing &mdash; it could be a service, a library,
+a binary, etc. A target has four _phases_ (all optional, but at least one must
+be present):
+
+-  Build
+-  Test
+-  Execute
```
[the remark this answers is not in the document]
amyhua: I think that phase-definitions, to specify side-effect-freeness, is nice.
However, it should be obvious to a user, without thinking too much, to determine which phase corresponds to which commands they'd like to store under a build target.

IMO I think either there should be _many_ phases, capturing all sorts of use cases, making it very clear which commands go into which phases, OR there should be very _few_ phases, making wrong choices harder to do, and bucketing examples amongst phases, very clearly.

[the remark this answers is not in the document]
amyhua: and to play devil's advocate: are we _sure_ that there are _no_ "build" commands, that would be BOTH fitting for the Build phase, _and_ side effect free?

What if a "build" command, that legitimately belongs in that bucket, does cause a side effect? What should a user do, in that case, for us?

TLDR are we sure these phases are "MECE"? ([MECE principle](https://en.wikipedia.org/wiki/MECE_principle))

docs/concepts.md
```diff
@@ -0,0 +1,56 @@
...
+-  Release
+
+Each phase, if present, has a list of one or more commands to run. The build
+phase can declare output files. Output files will be accessible to other targets
+that directly depend on this target.
```
glacials, line 24: (unrelated to Amy's above comments, just on the same line) `build` is explicitly allowed to output files, but `test` is not. What about code coverage output?

zombiezen:
Abandoning this PR for the time being until there's more internal alignment.

PR #213, opened 2020-10-12
Separate out container IP expansion from BuildData

zombiezen:
I broke a number of builds in v0.4 because the signature of the `BuildData.IP` method changed to include a `context.Context`. It was not obvious that this would cause a breakage.

To prevent this from occurring in the future, I've introduced a new type that is intentionally very limited in what it exports and prominently documented what can't change. I've introduced unit tests that ensure that such substitutions will continue to work.

I have surveyed build configurations run against our CI to confirm that no other symbols in `BuildData` are being used, just `.Containers.IP`.

Fixes ch-2741


glacials (approved):
Nice fix! Love the extra guardrails.

PR #217, opened 2020-10-20
Fix panic in remotebuild

zombiezen:
Fixes ch-2824


johnewart (approved):
Worked on my machine! :fire:

PR #219, opened 2020-10-21
Add convenience functions for common biome operations

zombiezen:
I've found these come up a lot as I've been porting buildpacks, but it's a large enough change to warrant its own PR.

Updates ch-2744

/cc @glacials


jamesnaftel (approved):
Very slick.  It is shaping up nicely!

PR #221, opened 2020-10-21
Move download functions into ybdata package

zombiezen:
Mostly a code motion, but also restructures the internals so there's only a single `open(2)` syscall. When I port the buildpacks, they will use the already-opened file rather than re-opening the path.

/cc @glacials


jamesnaftel (approved):
I like the truncate change.  Good stuff.

PR #222, opened 2020-10-22
Add ybdata.IsNotFound error predicate

zombiezen:
Needed for Ruby buildpack which peeks to see whether a binary exists.

Updates ch-2744


jamesnaftel (approved):
I like the change.  LGTM.

PR #225, opened 2020-10-27
Fail build if Ruby install fails

glacials:
Currently if the Ruby buildpack's Ruby install fails (e.g. because you specify a bad ruby version like `ruby:blah`), the build is still marked as successful. This branch changes the build to fail.


zombiezen:
I believe #223 will also have the same effect, which will be part of the 0.5 release. Do you need this cherry-picked to 0.4?

glacials:
Ah cool, wasn't sure. Nope, this is not blocking me, just something I noticed. Will close in favor of #223!

PR #231, opened 2020-11-05
Run Rust installer in buildpack

zombiezen:
Otherwise, builds will run but the standard library crate can't be found. QA'd by running against a real Rust project.


jamesnaftel (approved):
LGTM

PR #232, opened 2020-11-06
Add Docker biome implementation

zombiezen:
Fixes ch-2743

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,8 +22,17 @@ The format is based on [Keep a Changelog][], and this project adheres to
 
 ### Changed
 
+-  The Docker container for a build is entirely ephemeral: a new container will
+   be started for each target and the container will be stopped and removed
+   at the end of building the target. The contents of the build's
+   `HOME` directory and the package directory will persist between runs, but
+   all other changes will be lost, particularly packages installed
+   with `apt-get`.
+-  `yb build` now directly runs commands in the Docker container instead of
+   invoking a copy of itself to run the build.
 -  `yb run` now runs commands in the environment of a build target, not an
-   exec environment.
+   exec environment. This also means that `yb run` will operate in an ephemeral
+   Docker container by default.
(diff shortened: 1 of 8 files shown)


jamesnaftel (approved):
Woohoo!  LGTM.  Do you see having a ContainerOption for running as root vs regular user?

zombiezen:
> Do you see having a ContainerOption for running as root vs regular user?

Thanks for asking, I've been meaning to write this up on ch-1780. 🙂

So... overall yes, I would like to stop container builds from requiring elevated privileges. However, there's a substantial and reasonable use case of builds wanting to run `apt-get`, and I don't know that our buildpacks are well-equipped to satisfy those use cases. At least medium-term, we're probably stuck supporting root in builds.

However, supporting root in the build environment doesn't mean that it actually needs to be, well, root. Linux has a concept of a [user namespace](https://lwn.net/Articles/532593/), which permits you to remap the UIDs inside the container to a totally arbitrary set of UIDs on the host. However, this requires [some pretty significant hackery](https://docs.docker.com/engine/security/userns-remap/) to your friendly neighborhood Docker daemon, which I don't think is reasonable to ask the user. I wholeheartedly agree with the sentiment on docker/docker.github.io#7071 that "the system-wide mapping being the only option gives me goosebumps".

I propose to [do what Bazel does](https://blog.bazel.build/2015/09/11/sandboxing.html) and effectively create the container ourselves out of Linux namespaces. We can have yb pull Docker images directly from the registry, create and manage the isolated filesystem, and then arrange for the run commands to be in an appropriate user, mount, and PID namespace. This strategy wouldn't work for build targets that require side-containers (the networking would be hard to get right), but those would require running Docker anyway. What's nice about the biome-per-target arrangement is that yb can make that decision for the user dynamically: if the build target doesn't ask for Docker containers specifically, then we can use our lightweight, unprivileged containers.

This is a bigger undertaking than what I want to tackle in this yb release, so I'm punting to a post-0.5 release, since what's implemented with this biome is not strictly worse than what we have in 0.4.

PR #234, opened 2020-11-06
Delete old integration tests

zombiezen:
We're now getting higher confidence from unit tests and those run reliably on every commit. It's slow and hard to run these integration tests during development. I sampled a few to ensure they work before deleting.


jamesnaftel (approved):
LGTM.

PR #239, opened 2020-11-10
Automate Narwhal updates with dependabot

glacials:
Shortest interval is daily, so if we ever want a fast turnaround we'll still have to do it ourselves.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,7 @@ updates:
     schedule:
       interval: "daily"
     allow:
+      - dependency-name: "github.com/yourbase/commons"
       - dependency-name: "github.com/yourbase/narwhal"
     reviewers:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    allow:
+      - dependency-name: "github.com/yourbase/narwhal"
```
zombiezen, line 8:
```suggestion
      - dependency-name: "github.com/yourbase/narwhal"
      - dependency-name: "github.com/yourbase/commons"
```

zombiezen (approved):
LGTM. One optional suggestion:

PR #244, opened 2020-11-17
Decouple .yourbase.yml parse structures from execution data structures

zombiezen:
The execution data structure is easier for internal code to use and construct for test. For example, buildpacks overrides and target references are evaluated during parse.

I made YAML decoding strict, so unknown keys will now fail package loading. This works for most configurations I found used in production.

I removed the nonfunctional `exec.ports` key, which has never had a stable release where this works. Users should migrate to setting ports on `exec.container`.

Fixes ch-2285
Fixes ch-2550


zombiezen:
@jamesnaftel yb 0.5 seems stable enough that I'm ready for this to be reviewed.

jamesnaftel (approved):
LGTM.  I like the clarity.  Even simple things like changing from "Root" to "RunDir".

PR #257, opened 2020-11-30
Fix exec crash when default environment is empty

zombiezen:
Fixes ch-3170

### DIFF
--- a/cmd/yb/exec.go
+++ b/cmd/yb/exec.go
@@ -113,6 +113,9 @@ func (b *execCmd) run(ctx context.Context) error {
 		if err != nil {
 			return err
 		}
+		if len(overrideEnv) > 0 && deps.EnvironmentTemplate == nil {
+			deps.EnvironmentTemplate = make(map[string]string)
+		}
 		for k, v := range overrideEnv {
 			deps.EnvironmentTemplate[k] = v
 		}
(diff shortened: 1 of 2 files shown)


cmd/yb/exec.go
```diff
@@ -113,6 +113,9 @@ func (b *execCmd) run(ctx context.Context) error {
 		if err != nil {
 			return err
 		}
+		if len(overrideEnv) > 0 && deps.EnvironmentTemplate == nil {
+			deps.EnvironmentTemplate = make(map[string]string)
+		}
```
glacials, line 118: This feels smelly. I'm not confident that we won't hit this same issue later in some other code path.

1. If we are passing `MapVars` an empty slice on line 95 (I'm not sure if it's empty or nil) it feels like it should respond with an empty map. Its current behavior is to respond with `nil`.
2. From a product perspective an empty environment and a nil environment feel identical in effect, so it's unclear to me what we are gaining by keeping the possibility of nil environment maps.

glacials (changes requested):
Thanks for the fix! Just one comment, but if this is just a hotfix and will be overwritten by #244, I'm happy to withdraw it.

cmd/yb/exec.go
zombiezen: This is a hotfix that will be obsoleted by #244. I have the same qualms with this code.

glacials: then lgtm!

glacials (approved):

PR #258, opened 2020-12-01
Cancel command context on SIGINT/SIGTERM

zombiezen:
The Docker biome reacts by restarting the container. This stops any currently running processes and then permits any final cleanup to run at the expense of any concurrent processes.

Included a small fix to ensure buildpack archive extraction cleanup does not receive an interrupt.

Fixes ch-2213

### DIFF
--- a/internal/biome/docker.go
+++ b/internal/biome/docker.go
@@ -253,6 +253,35 @@ func (c *Container) Run(ctx context.Context, invoke *Invocation) error {
 		return fmt.Errorf("run in container %s: %w", c.id, err)
 	}
 
+	// Unfortunately, there's no way to signal a particular exec:
+	// https://github.com/moby/moby/issues/9098
+	//
+	// So if the Context is cancelled, we restart the container, which may take
+	// down other running processes, but usually the biome is going down anyway.
+	runDone := make(chan struct{})
+	interrupterDone := make(chan struct{})
+	go func() {
+		defer close(interrupterDone)
+		select {
+		case <-ctx.Done():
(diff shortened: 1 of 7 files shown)


internal/biome/docker.go
```diff
@@ -253,6 +253,35 @@ func (c *Container) Run(ctx context.Context, invoke *Invocation) error {
...
+				&docker.HostConfig{},
+				xcontext.IgnoreDeadline(ctx),
+			)
+			if err != nil {
+				log.Warnf(ctx, "Could not restart container %s after interrupt: %v", c.id, err)
+				return
+			}
+		case <-runDone:
+		}
+	}()
```
glacials, line 283: I'm confused about why this works. I see the GitHub issue that we can't signal a specific `docker exec`, so a duct tape restart makes sense in the context of "we are trying to kill just the process started by `docker exec`".

But I don't follow how the restarting of the container is avoided during a Ctrl-C on `yb exec`. Or is it restarted, then immediately stopped by a different code path?

glacials (approved):
Tested locally and seems to solve ch-2213 for me 😄 Thanks!

internal/biome/docker.go
zombiezen: It is the latter: it is restarted and then cleaned up later by `*Docker.Close`. The idea is to keep the Docker container running in steady-state until the caller closes the biome.

This change does not handle concurrent `Run`s well, because the stops and starts could race with each other. A better implementation would coalesce and synchronize such signals. However, this doesn't become a real issue until we try to implement ch-1251, so adding a note there.

PR #275, opened 2021-01-25
Set user.home system property in Java buildpack

zombiezen:
Causes the Maven dependency cache to be stored in the proper directory.

Fixes ch-3780


jamesnaftel (approved):
LGTM.  Thx!

PR #281, opened 2021-02-25
Set locale environment variables in build environment

zombiezen:
I selected `C.UTF-8` as a somewhat neutral default that is likely to be available in many environments. The user can always override it in their target configuration.

Also corrected the non-POSIX-compliant `TZ` value from `UTC` to `UTC0`.

I added a small companion document to specify the build environment and what guarantees we provide going forward.

Fixes ch-3824


clausti (approved):
🚀

PR #282, opened 2021-02-25
Add new default "prefer host" mode

zombiezen:
Runs targets on the host, but spins up resource containers if the target requires them. Added a new `--mode` option to handle the three states. If a target specifies container information, then it will be run inside a Docker container.

Fixes ch-3867

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,12 +10,27 @@ The format is based on [Keep a Changelog][], and this project adheres to
 
 ### Added
 
+-  A new `--mode` option for `build`, `exec`, and `run` allows specifying
+   whether commands should be run inside or outside Docker.
 -  yb now obeys the `DOCKER_HOST` environment variable.
+
+### Changed
+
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,26 @@ The format is based on [Keep a Changelog][], and this project adheres to
...
+
+### Added
+
+-  A new `--mode` option for `build`, `exec`, and `run` allows specifying
+   whether commands should be run inside or outside Docker.
+
+### Changed
+
+-  Commands run as part of `build`, `exec`, or `run` now run without Docker by
+   default. You can get the old behavior by running with `--mode=container`.
```
jamesnaftel, line 19: 👍

jamesnaftel (approved):
Good stuff.

zombiezen:
I'm pretty excited about this. I want to wait until ch-3426 is resolved before moving this forward, but from my tests so far, this gives an all-around better out-of-the-box experience with yb.

PR #285, opened 2021-03-23
Port content from docs.yourbase.io

zombiezen:
Most of the content is copied verbatim, with some light reformatting for GitHub-Flavored Markdown. However, the List of Build Packs was re-authored with the most up-to-date information and now includes examples of accepted versions.

I omitted "Getting Started with YourBase" and "How to Write Build Targets" because they are redundant with the content in README.md. The "Remote Building" page is better served by `yb remotebuild --help` instead of having a separate page we have to update.

Fixes ch-4338

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,7 +177,7 @@ yb run --target=foo bash
 
 ## Documentation
 
-More documentation available at http://docs.yourbase.io
+More documentation available in the [docs folder](docs).
 
 ## Contributing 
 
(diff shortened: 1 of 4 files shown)


zombiezen:
Hey @amyhua, do you have bandwidth to review this or should I send to someone else? I'd like to be able to link to these docs without using a PR-specific permalink. Thanks!

amyhua (approved):

PR #286, opened 2021-03-30
Fix `yb init` crash if Docker is not working

zombiezen:
A superfluous error check causes `yb init` to bail before project language detection if there was a problem connecting to the Docker daemon.


zombiezen:
Gentle ping, @jamesnaftel

jamesnaftel (approved):

PR #289, opened 2021-04-14
Fix installation and getting started links

jamesnaftel:


README.md
```diff
@@ -32,149 +32,6 @@ The primary features of the YB tooling are:
 
 ![magic!](http://www.reactiongifs.com/r/mgc.gif)
```
zombiezen, line 34: Let's keep the section headings with a small blurb directing folks to the new pages.

```suggestion
## Installation

yb is available for Mac via Homebrew, and Linux/WSL2 via `apt-get`.
Installation instructions are available at https://docs.yourbase.io/installation.html.

## Getting Started

To use yb, you need a `.yourbase.yml` file at the top of your project directory.
You can run `yb init` to generate one:

```shell
cd path/to/my/project
yb init
yb build
\```

For a more in-depth tutorial, see https://docs.yourbase.io/getting-started.html
```

zombiezen (changes requested):
Also, we should change the redirect under docs/build/getting_started.html to point to https://docs.yourbase.io/getting-started.html

zombiezen (approved):

PR #291, opened 2021-04-16
Upgrade to Go 1.16

zombiezen:
Move `yb init` templates into separate files using the embedding feature of Go 1.16.


jamesnaftel (approved):
Neat.  Had to read about embed.FS

PR #293, opened 2021-04-21
Don't create Docker network if container biomes not required

zombiezen:
Also fix receiver of `executionMode.Set` causing `--mode` flag to no-op.

Fixes ch-3867


jamesnaftel (approved):
Love it!

PR #295, opened 2021-04-22
Make output more compact and easier to read

zombiezen:
Build output now includes a target and command heading. Info-level log lines no longer have the unhelpful "INFO" prefix. All non-error build output now goes to stdout so someone who wants a more quiet experience for a successful build can redirect to `/dev/null`.

Individual lines of command and log output now also include the target they are running for and the time the line was outputted. This serves to both visually group outputs by command as well as provide fine-grained timing information.

Fixes ch-4043
Fixes ch-4040

Terminal output for a successful build:

Screenshot of terminal output from a successful build

Terminal output for a failed build:

Screenshot of terminal output from a failed build

For comparison, output from the same build on yb v0.6.3:

Screenshot of terminal output from a successful build in yb v0.6.3. It is longer and less organized.

FYI @amyhua

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ The format is based on [Keep a Changelog][], and this project adheres to
 
 -  Commands run as part of `build`, `exec`, or `run` now run without Docker by
    default. You can get the old behavior by running with `--mode=container`.
+-  Tool output has been changed to be more compact, to be easier to trace
+   command output, and to include more timing information.
 -  `yb platform` is now an alias for `yb version`.
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ The format is based on [Keep a Changelog][], and this project adheres to
 -  A new `--mode` option for `build`, `exec`, and `run` allows specifying
    whether commands should be run inside or outside Docker.
 -  yb now obeys the `DOCKER_HOST` environment variable.
+-  yb now obeys the [`NO_COLOR` environment variable][] and propagates it to the
+   build environment.
+
+[`NO_COLOR` environment variable]: https://no-color.org/
 
 ### Changed
 
--- a/cmd/yb/build.go
+++ b/cmd/yb/build.go
@@ -148,7 +148,7 @@ func (b *buildCmd) run(ctx context.Context) error {
 		return alreadyLoggedError{buildError}
 	}
 
-	fmt.Printf("%sBUILD SUCCEEDED%s ️✔️\n", style.buildResult(true), style.reset())
+	fmt.Printf("%sBUILD PASSED%s ️✔️\n", style.buildResult(true), style.reset())
 	return nil
 }
 
--- a/cmd/yb/build.go
+++ b/cmd/yb/build.go
@@ -3,7 +3,7 @@ package main
 import (
 	"context"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"os"
 	"strings"
 	"time"
@@ -99,6 +99,7 @@ func (b *buildCmd) run(ctx context.Context, buildTargetName string) error {
 	startTime := time.Now()
 	ctx, span := ybtrace.Start(ctx, "Build", trace.WithNewRoot())
 	defer span.End()
+	ctx = withStdoutLogs(ctx)
 
 	log.Infof(ctx, "Build started at %s", startTime.Format(longTimeFormat))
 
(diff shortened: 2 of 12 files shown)


amyhua:
Love this!! <3 It is less noisy and much clearer when a build fails, now. I'm glad we're taking steps to clean up the build output now.

Some suggestions:

* I may suggest that the build success message (-- BUILD SUCCEEDED --) could be formatted to look more like the build failure message (BUILD FAILED X in _red_ --> BUILD PASSED ✓ in _green_) for visual consistency. I think the green coloring on a passed build message has a nice feel-good sensation to it. :) I also think the word "passed" is more standard and better to read than "succeeded".
* Also, is the "Elapsed" header supposed to be closer to the point where the text starts? At the starting char of "0.838s...". In the screenshot at least, the vertical alignment of the Elapsed title as well as its content seem to be vertically off.
* Nitpick: omit "exit code 1"? since that is not very informative, and appears with every failed build.

---

Perhaps as part of this work or for a future iteration, we can also consider whether or not it is necessary to group the information by "Target: default" if there is only 1 target, with no name. For the single-default-target case, the logs could be even cleaner looking, with just

```
15:55:48 | Action taken message here...
15:55:49 | Action taken message here...
15:55:52 | Action taken message here...
```

Since the target-grouping is only valuable to organize by when there is more than 1 target, and you are looking to distinguish actions taken by various targets.

---

For multi-line output, I would suggest

```
15:55:48 | Action taken message here that goes on...
         | for multiple lines...
         | ending here
15:55:52 | Another action taken message here...
```
for vertical alignment of the description, keeping the timestamp / metadata info always to the left in the layout of information.

jamesnaftel (approved):
LGTM.  Love the new output!

zombiezen:
@amyhua:
> Love this!! <3 It is less noisy and much clearer when a build fails, now. I'm glad we're taking steps to clean up the build output now.

:tada:

> Some suggestions:
> 
> * I may suggest that the build success message (-- BUILD SUCCEEDED --) could be formatted to look more like the build failure message (BUILD FAILED X in _red_ --> BUILD PASSED ✓ in _green_) for visual consistency. I think the green coloring on a passed build message has a nice feel-good sensation to it. :) I also think the word "passed" is more standard and better to read than "succeeded".

I've been going back and forth on this. I'll give you my rationale for the current output, and if you still want the change, I'll go ahead and make the changes.

- "Tests passed" and "tests failed" are the two common verbs, but yb targets aren't always running tests. For Go, the `default` target very commonly builds the binary. "Build succeeded" sounds like the more correct verb to my ears.
- I put errors messages in red because that is the typical convention, but success for most CLIs do not usually have any additional coloring (see `git fetch`, `git clone`, `npm install`). While green definitely stands out for black or white default text color, green-on-black is another somewhat typical color scheme.

> * Also, is the "Elapsed" header supposed to be closer to the point where the text starts? At the starting char of "0.838s...". In the screenshot at least, the vertical alignment of the Elapsed title as well as its content seem to be vertically off.

Done!

> * Nitpick: omit "exit code 1"? since that is not very informative, and appears with every failed build.

Whoops, sorry I should have clipped the screenshot before that line. That's an artifact of my zsh setup, not output from the tool itself.

> For the single-default-target case, the logs could be even cleaner looking, with just
> 
> ```
> 15:55:48 | Action taken message here...
> 15:55:49 | Action taken message here...
> 15:55:52 | Action taken message here...
> ```
> 
> The by-target-grouping is only valuable to organize by-target when there is more than 1 target, and a user is looking to distinguish actions taken by various targets.
> 
> For multi-line output, I would suggest
> 
> ```
> 15:55:48 | Action taken message here that goes on...
>          | for multiple lines...
>          | ending here
> 15:55:52 | Another action taken message here...
> ```
> 
> for vertical alignment of the description, keeping the timestamp / metadata info always to the left in the layout of information.

Good idea! It requires a more stateful logger than we have now, which bumps up scope. Since this PR is already pretty far-reaching, I'd like to address in a separate PR. I've opened up ch-4629 to track the follow-up.

amyhua:
@zombiezen looking good. I think the only outstanding issue is about build passed vs build succeeded, and picking the right verb + decide on coloring or not.

I agree, builds don't really "pass" as they don't have tests all the time that pass. A build "passing" is like viewing the whole build as having a lack of failures, and successfully finished.
I looked at other CIs and it looks like:
* TravisCI displays a "passed" status for succeeded builds
* CircleCI uses build "failed or passed"
* Jenkins does use "build succeeded"
* In Azure, it is "succeeded"
I've been used to using CirlceCI in the past and was partial to "passed". I liked that "passed" is the same character length as "failed" because it eventually makes visual design around both of them look more consistent, haha. But it's not a big deal either way, feel free to choose what feels right.

Format wise, would be nice to be consistent in either way: "BUILD [SUCCEEDED/PASSED] ✓" to match the "BUILD FAILED X" format and look.

Re: coloring -- at least on UIs, shiny green indicators for success gives the user a nice feel-good kick out of finishing something, but I haven't been sure about the outstanding convention on CLI design.

zombiezen:
Sounds like "BUILD PASSED ✓" has enough convention going for it. I've gone ahead and made that change.

My intuition is to keep the default foreground color (instead of forcing green) for "BUILD PASSED ✓", try it out within the team, and see how we feel.

PR #296, opened 2021-04-22
First draft of ruby instructions

jamesnaftel:
Here is the first draft.  There are a couple of things (like the token request) that I expect to change.

Expected changes:
* Need to request a token
* YB_COHORT_COUNT, YB_TEST_COHORT, rename to "YOURBASE_"
* Add s3 bucket secret support

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -8,3 +8,7 @@
 - [Package Configuration Reference](configuration.md)
 - [CI Acceleration and Caching](ci-caching.md)
 - [Design Philosophy](design-philosophy.md)
+
+# Quick Start Acceleration
+
+- [YourBase Ruby Acceleration](yourbase-ruby.md)
--- a/docs/yourbase-ruby.md
+++ b/docs/yourbase-ruby.md
@@ -1,0 +1,59 @@
+# YourBase Ruby Acceleration
+
+Tests are important. For large monoliths, they're also a major source of drag on velocity.
+
+YourBase is a tool that traces your tests to determine which functions each test depends on. It can later use this information to determine which tests do not need to run because their code paths have not changed. These tests are skipped automatically.
+
+YourBase works with Ruby versions >= 2.3
+
+## Getting Started
+No configuration, setup, or babysitting required. To get started, you need the YourBase gem and a download token.  To request a token, please visit [YourBase.io](https://yourbase.io/download)
+
+Once you have a token, simply follow the steps below:
+```sh
+YOURBASE_DOWNLOAD_TOKEN=<enter token> 
+bundle add yourbase-rspec --git "https://$YOURBASE_DOWNLOAD_TOKEN:x-oauth-basic@github.com/yourbase/yourbase-rspec-skipper-engine.git"
+
+bundle install
+```
+
+## First run
+After installing yourbase-rspec, if you are not using Rails you must add
+"require 'yourbase-rspec'" in your spec folder.
+
+```sh
+# Add require for non Rails projects
+echo "require 'yourbase-rspec'" >> spec/yourbase_spec.rb
+```
+
+Run your tests with the same command you typically use. You should see a rocket ship at the beginning the RSpec test section.
+
+```
+🚀
+```
+
+The first run will be cold, so if you just want to see YourBase in action and your tests are going to take a while, you can run a subset of tests. Tracing data for the subset will be used correctly even if you later run all tests.
+
+After the run finishes, running again will skip all tests. Modifying a dependency will run only tests whose code paths touched the changed code. You're YourBased! 🚀
+
+## RSpec Output
+
+YourBase enhances the output so that you can clearly see the results of the Gem.
+
+For an accelerated run, you will see the number of skipped tested added to your
+RSpec summary line:
+```sh
+1 examples, 0 failures, 1 skipped with YourBase🚀
+```
+
+To get additional details, please use -f, --formatter [p]rogress | [d]ocumentation]
+
+## Cohorting / Sharding
+YourBase supports sharding your tests without negatively affecting tracing or acceleration. It uses consistent hashing to split tests into cohorts that stay the same between runs, even as the test pool grows or shrinks.
+
+1) Set YB_COHORT_COUNT to the number of cohorts your tests should be split into. This should be the same among all shards.
+1) Set YB_TEST_COHORT to the cohort ID this run should identify as, starting with 1. This should be different among all shards.
+Without these set, YourBase assumes a value of "1" for both, meaning one shard will run one cohort; that cohort will contain all tests.
+
+Note that tests are only guaranteed to remain in the same cohort as long as
+YB_COHORT_COUNT doesn't change.


docs/README.md
```diff
@@ -8,3 +8,7 @@
 - [Package Configuration Reference](configuration.md)
 - [CI Acceleration and Caching](ci-caching.md)
 - [Design Philosophy](design-philosophy.md)
+
+# Quick Start Acceleration
```
zombiezen, line 12: I'd recommend putting this at the top: before the CI and CLI section.

docs/yourbase-ruby.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Getting Started
+No configuration, setup, or babysitting required. To get started, you need the YourBase gem and a download token.  To request a token, please visit [YourBase.io](https://yourbase.io/download)
+
+Once you have a token, simply follow the steps below:
+```sh
+YOURBASE_DOWNLOAD_TOKEN=<enter token> 
+bundle add yourbase-rspec --git "https://$YOURBASE_DOWNLOAD_TOKEN:x-oauth-basic@github.com/yourbase/yourbase-rspec-skipper-engine.git"
+
+bundle install
```
zombiezen, line 17: I recommend following the advice in https://www.vidarholen.net/contents/blog/?p=958:

```suggestion
bundle add yourbase-rspec --git "https://${YOURBASE_DOWNLOAD_TOKEN?}:x-oauth-basic@github.com/yourbase/yourbase-rspec-skipper-engine.git" &&
bundle install
```

docs/yourbase-ruby.md
```diff
@@ -0,0 +1,59 @@
...
+
+After the run finishes, running again will skip all tests. Modifying a dependency will run only tests whose code paths touched the changed code. You're YourBased! 🚀
+
+## RSpec Output
+
+YourBase enhances the output so that you can clearly see the results of the Gem.
+
+For an accelerated run, you will see the number of skipped tested added to your
+RSpec summary line:
+```sh
```
zombiezen, line 45:
```suggestion
```plain
```

docs/yourbase-ruby.md
```diff
@@ -0,0 +1,59 @@
...
+1 examples, 0 failures, 1 skipped with YourBase🚀
+```
+
+To get additional details, please use -f, --formatter [p]rogress | [d]ocumentation]
```
zombiezen, line 49: This seems like it's supposed to include a documentation link?

zombiezen (commented):
Looks like a good start to me. I had a few suggestions on organization and formatting, but I will defer to @clausti for technical accuracy.

docs/README.md
jamesnaftel: For Python,  do you think we should link to pypi yourbase from here?

docs/yourbase-ruby.md
jamesnaftel: I like this!  Very clear.

docs/README.md
zombiezen: Yes! 👍

docs/yourbase-ruby.md
```diff
@@ -0,0 +1,73 @@
+# YourBase Ruby Acceleration
+
+Tests are important. For large monoliths, they're also a major source of drag on velocity.
+
+YourBase is a tool that traces your tests to determine which functions each test depends on. It can later use this information to determine which tests do not need to run because their code paths have not changed. These tests are skipped automatically.
```
clausti, line 5:
```suggestion
YourBase is a tool that traces your tests to determine which files they depend on. It uses this information to determine which tests must run, because they are new or have changed dependencies. Tests without changed dependencies are skipped.
```

docs/yourbase-ruby.md
```diff
@@ -0,0 +1,73 @@
...
+
+YourBase works with Ruby versions >= 2.3
```
clausti, line 7:
```suggestion
YourBase works with Ruby versions >= 2.3 and RSpec 3+
```

docs/yourbase-ruby.md
```diff
@@ -0,0 +1,73 @@
...
+After installing yourbase-rspec, if you are not using Rails you must add
+"require 'yourbase-rspec'" in your spec folder.
+
+```sh
+# Add require for non Rails projects
+echo "require 'yourbase-rspec'" >> spec/yourbase_spec.rb
+```
+
+Run your tests with the same command you typically use. You should see a rocket ship at the beginning the RSpec test section.
+
```
clausti, line 27: We should have a section about Spring here, because it's the most common reason they won't see the rocketship. 

Right now, they need to run `spring stop` from the command line to turn it off, but it would be better if we could figure out if manually requiring `yourbase-rspec` at the top of the spec_helper.rb gets around the incompatibility, so that we could tell them to do that instead of having to stop Spring. Stopping Spring is definitely fine in the CI but is a little kludgier for local dev where Spring does actual useful things, including shorten the spin-up time for "loading the rails environment" for specs.

docs/yourbase-ruby.md
```diff
@@ -0,0 +1,73 @@
...
+🚀
+```
+
+The first run will be cold, so if you just want to see YourBase in action and your tests are going to take a while, you can run a subset of tests. Tracing data for the subset will be used correctly even if you later run all tests.
+
+After the run finishes, running again will skip all tests. Modifying a dependency will run only tests whose code paths touched the changed code. You're YourBased! 🚀
+
+## RSpec Output
+
+YourBase enhances the output so that you can clearly see the results of the Gem.
```
clausti, line 38:
```suggestion
YourBase adds to the RSpec output to give you information while about if examples are being run or skipped, and why. 
```

docs/yourbase-ruby.md
```diff
@@ -0,0 +1,73 @@
...
+
+For an accelerated run, you will see the number of skipped tested added to your
+RSpec summary line:
+```plain
+1 examples, 0 failures, 1 skipped with YourBase🚀
+```
```
clausti, line 44:
```suggestion
The default RSpec output from the `ProgressFormatter` (`.....*..... F........`) will print a `.` in yellow (or the color you have set for `:pending`) when an example is skipped. 

The `DocumentationFormatter` will add the reason an example group is being run in green (or the color your have set for `:success`), and the reason an example group is being skipped in yellow (or the color you have set for `:pending`). 

The summary line will show how show how many examples were skipped! 🚀
```plain
1 examples, 0 failures, 1 skipped with YourBase🚀
```
```

File: docs/yourbase-ruby.md [hunk 10]
```diff
@@ -0,0 +1,73 @@
...
+
+YourBase adds additional details when using the RSpec formatter option with the
+progress and documentation formatters.
```
clausti, line 47:
```suggestion

```

clausti (commented):
Gotta go to a scheduled thing but will come back to this in a couple hours.

docs/yourbase-ruby.md
jamesnaftel: Thx, copy and paste from Python was a bad idea here.  :)

docs/yourbase-ruby.md
```diff
@@ -0,0 +1,73 @@
...
+bundle add yourbase-rspec --git "https://${YOURBASE_DOWNLOAD_TOKEN?}:x-oauth-basic@github.com/yourbase/yourbase-rspec-skipper-engine.git" && bundle install
+```
+
+## First run
+After installing yourbase-rspec, if you are not using Rails you must add
+"require 'yourbase-rspec'" in your spec folder.
+
+```sh
+# Add require for non Rails projects
+echo "require 'yourbase-rspec'" >> spec/yourbase_spec.rb
```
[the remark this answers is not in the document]
jamesnaftel: Let's chat.  I wanted to give an example of a simple, quick way to add in our requirement without touching their files.

docs/yourbase-ruby.md
jamesnaftel: Ok.  Have this in my notes but wanted to try it out first.  Will add it.

jamesnaftel:
Thx for the feedback @clausti!

jamesnaftel:
For the email, I added "support@yourbase.io".  We also have "ruby@yourbase.io".  I was torn between the two.  Happy to discuss.

docs/yourbase-ruby.md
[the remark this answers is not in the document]
clausti: we can chat at/after standup, but they already have to change a project file to add the gem. And looks like what this will do is, assuming they are at the root dir of the project, add a file named yourbase_spec.rb to the spec folder that only contains the require line? I don't think that will get picked up if they run a specific spec file from their project.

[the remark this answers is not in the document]
clausti: I think this `echo` line needs to be removed before we publish this. It will stop a nontrival percentage of rails devs in their tracks as incomprehensible or exceptionally odd.

```suggestion
If you are not in a Rails project, you will also need to add `require 'yourbase-rspec'` to your `spec_helper.rb` (or another project file which will be loaded, such as the file containing the specs you are running).
```

docs/yourbase-ruby.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+Run your tests with the same command you typically use. You should see a rocket ship at the beginning the RSpec test section.
```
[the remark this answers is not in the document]
clausti: If you would not like to take the change verbatim, can you update the text to include the information about compatibility with command line arguments and the literal emoji? It's not clear to me that everyone (especially devs with a different native language) will automatically know "rocketship" means "🚀"

docs/yourbase-ruby.md
clausti: I think for now it is sufficient to say that if you do not see the rocketship but ARE running Spring, that it should be turned off to accelerate the tests.

docs/yourbase-ruby.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Getting Started
+1. Add `gem 'yourbase-rspec', '~> 0.5.6',` to your Gemfile. If you have a `:test` group, add it there.
+2. `bundle install` from the command line.
+3. If you are not in a Rails project, you will also need to `require 'yourbase-rspec` your spec_helper (or at the top of the spec file you want to run).
+
+Once you have a token, simply follow the steps below:
+```sh
+bundle add yourbase-rspec --git "https://${YOURBASE_DOWNLOAD_TOKEN?}:x-oauth-basic@github.com/yourbase/yourbase-rspec-skipper-engine.git" && bundle install
+```
```
clausti, line 17: sorry if I missed this in the first pass, but I think the 1,2,3 and the ```sh``` section below it assume different things about the the public/privateness of the repo.

If we're going to publish this today-ish for Gusto purposes, we should make step `1.` include the token syntax,
`gem 'yourbase-rspec', '~> 0.6.0, git: "https://yourbase-tokenbot:#{ENV.fetch(YOURBASE_DOWNLOAD_TOKEN){nil}}@github.com/yourbase/yourbase-rspec-skipper-engine", branch: 'main'`, and then lines 14-16 are redundant and can be removed.

PR #299, opened 2021-04-23
Show warnings if user is not in Docker group

zombiezen:
Refactored helper code to ensure there is a single codepath for determining whether to use Docker.

Fixes ch-2945


zombiezen:
FYI @glacials, since you originally requested this feature.

jamesnaftel (approved):
LGTM.  Naming changes are more clear as well.

PR #300, opened 2021-04-23
Use topological sort to compute build order

zombiezen:
With this comes the ability to specify multiple targets on the command line and have them run in the correct order.

Fixes ch-2750


jamesnaftel (approved):
LGTM.

PR #303, opened 2021-05-20
Fill out package configuration reference

zombiezen:
Add missing documentation about containers, build packs, and exec.  Included some new examples, including best practices for using apt.

Fixes ch-4759
Fixes ch-1426

PR #305, opened 2021-05-20
Permit environment sections to be maps instead of lists

zombiezen:
Lists are still permitted for backward compatibility. I've updated the docs to use the new map style, as it's a better fit for YAML structures.

Fixes ch-4916

~~**Note to reviewers:** I'm sending this out as draft because it modifies the docs I introduced in #303. Once that PR lands, I'll mark this as ready for review.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ The format is based on [Keep a Changelog][], and this project adheres to
 -  A new `--mode` option for `build`, `exec`, and `run` allows specifying
    whether commands should be run inside or outside Docker.
 -  `yb build` can now build multiple targets in one invocation.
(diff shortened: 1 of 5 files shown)


jamesnaftel (approved):
LGTM.  I did not know about the Unmarshaler interface.  Had to look it up.  :)

zombiezen:
> I did not know about the Unmarshaler interface. Had to look it up. :)

Yeah, same here: most other Unmarshaler interfaces take in bytes (see [`BinaryUnmarshaler`](https://pkg.go.dev/encoding#BinaryUnmarshaler) and [`json.Unmarshaler`](https://pkg.go.dev/encoding/json#Unmarshaler)), so this one's a little unique.

Added a doc comment, since it is a little obscure.

PR #306, opened 2021-05-20
Mark Docker as necessary if starting side containers

zombiezen:
There are two distinct use cases for querying whether a list of targets will use Docker: biome creation and Docker initialization. The former only cares about whether the target specifies a container, but the latter also cares about side containers. I've split the check into two separate functions to make the logic clear and allow easy testing.

This is a bug present in 0.7.0-alpha1, but it is not present in 0.6.  No backporting is necessary.


jamesnaftel (approved):
LGTM

PR #309, opened 2021-05-26
Add some bits to support installing Ruby properly on Apple M1 

johnewart:
More fixes coming soon; this lets Ruby 2.7.x and earlier compile correctly on M1


zombiezen:
Would you mind running the following on your M1 machine:

```shell
cd internal/buildpack && go test -run=TestRuby -record
```

The test failure is because the `darwin/amd64` test wasn't re-recorded and it's detecting that the set of commands has changed. I can re-generate that file for you on my Mac mini.

PR #314, opened 2021-06-01
Add support for SBT

johnewart:
Also support `.tgz` and `.tbz2` files in decompression bits

### DIFF
--- a/internal/buildpack/sbt.go
+++ b/internal/buildpack/sbt.go
@@ -1,0 +1,111 @@
+// Copyright 2021 YourBase Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package buildpack
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"os"
+	"io/ioutil"
+
+	"github.com/yourbase/yb"
+	"github.com/yourbase/yb/internal/biome"
+	"zombiezen.com/go/log"
+)
+
+func installSBT(ctx context.Context, sys Sys, spec yb.BuildpackSpec) (_ biome.Environment, err error) {
+	installDir := sys.Biome.JoinPath(sys.Biome.Dirs().Tools, "sbt", spec.Version())
+	desc := sys.Biome.Describe()
+	parentDir := os.TempDir()
+	socketDir, err := ioutil.TempDir(parentDir, "*-ybsbt")
+	if err != nil {
+		log.Errorf(ctx, "Error creating directory for SBT sockets: %v", err)
+	}
+	log.Infof(ctx, "SBT will use %s as the socket directory", socketDir)
+//	defer os.RemoveAll(socketDir) // clean up
+
+	env := biome.Environment{
+		Vars: map[string]string{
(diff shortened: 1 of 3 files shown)


internal/buildpack/sbt_test.go
```diff
@@ -0,0 +1,46 @@
...
+		Stderr: versionOutput,
+	})
+	t.Logf("sbt --version output:\n%s", versionOutput)
+	if err != nil {
+		t.Errorf("sbt --version: %v", err)
+	}
+	if got := versionOutput.String(); !strings.Contains(got, version) {
+		t.Errorf("sbt --version output does not include %q", version)
+	}
+}
```
zombiezen, line 46: It would be good to have some unit tests for the `sbtDownloadURL` function, like for the Java buildpack (but doesn't need to be nearly as extensive): https://github.com/yourbase/yb/blob/03eed7bdf70d1546b07a3490942abc8d78b78da4/internal/buildpack/openjdk_test.go#L48-L141

internal/buildpack/sbt_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"testing"
+
+	"github.com/yourbase/yb/internal/biome"
+	"zombiezen.com/go/log/testlog"
+)
+
+func TestSBT(t *testing.T) {
+	const version = "1.5.2"
+	ctx := testlog.WithTB(context.Background(), t)
+	// TODO(johnewart): SBT depends on Java  -- see Ross's comment in sbt tests
```
zombiezen, line 31:
```suggestion
	// TODO(johnewart): SBT depends on Java  -- see Ross's comment in Ant tests
```

internal/buildpack/sbt.go
```diff
@@ -0,0 +1,111 @@
...
+	}[desc.OS]
+	if data.OS == "" {
+		return "", fmt.Errorf("unsupported os %s", desc.OS)
+	}
+	data.MajorVersion = majorVersion
+	data.MinorVersion = minorVersion
+	data.PatchVersion = patchVersion
+	data.SubVersion = subVersion
+	return templateToString(urlPattern, data)
+}
```
zombiezen, line 111: IIUC, the URL is only dependent on version, right? I think this function can be simplified to pass the version string through verbatim.

```suggestion
func sbtDownloadURL(version string) (string, error) {
	const urlPattern = "https://github.com/sbt/sbt/releases/download/v{{ .Version }}/sbt-{{ .Version }}.tgz"

	var data struct {
		Version string
	}
	data.Version = url.PathEscape(version)
	return templateToString(urlPattern, data)
}
```

internal/buildpack/sbt.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/yourbase/yb"
+	"github.com/yourbase/yb/internal/biome"
+	"zombiezen.com/go/log"
+)
+
+func installSBT(ctx context.Context, sys Sys, spec yb.BuildpackSpec) (_ biome.Environment, err error) {
+	installDir := sys.Biome.JoinPath(sys.Biome.Dirs().Tools, "sbt", spec.Version())
+	desc := sys.Biome.Describe()
+	parentDir := os.TempDir()
+	socketDir, err := ioutil.TempDir(parentDir, "*-ybsbt")
```
zombiezen, line 35: This unfortunately won't work on Docker-based builds, since it creates the directory on the host instead of inside the biome.

Would it be appropriate to use a directory inside the biome home? Perhaps `sys.Biome.JoinPath(sys.Biome.Dirs().Tools, "sbt-server")`?

internal/buildpack/sbt.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		log.Errorf(ctx, "Error creating directory for SBT sockets: %v", err)
+	}
+	log.Infof(ctx, "SBT will use %s as the socket directory", socketDir)
```
zombiezen, line 39: I don't think the average user needs to know this information unless something went wrong, so I suggest debug-level log.

```suggestion
	log.Debugf(ctx, "SBT will use %s as the socket directory", socketDir)
```

PR #320, opened 2021-06-22
added message about disabling yourbase for debugging

jamesnaftel:
@glacials Do you think this rises to the level of being on the main page?  I considered adding a new section and blurp below "### For use in a CI".

I do not want to call it a known issue as that sounds like something we may need to "fix".

### DIFF
--- a/docs/python/configuration.md
+++ b/docs/python/configuration.md
@@ -180,6 +180,10 @@ echo ${XDG_STATE_HOME-~/.local/state}/yourbase/python.log
 
 When on, YourBase will not load.
 
+This setting is recommended when using a debugger.  Since YourBase Test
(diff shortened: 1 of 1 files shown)


glacials:
Is it easy & reasonable to do this detection in the library, like figure out when we are running under a debugger and print a warning with this information if so?

If not, I think this should indeed be in Known Issues, only because that's where people will look when they experience this behavior. Making it easy to figure out why we are behaving oddly outweighs the possibility that putting it there makes it look like something that needs fixing IMO.

jamesnaftel:
Sounds good.  I'll work on seeing if I'm in a debugger in the morning.  I see a couple of hits with a quick search.  If that does not work, I'll add it to known issues as well.

